PR #6, opened 2024-05-14
Combine PREMIS files from Vecteur AIPs

mcantelon:


cmd/worker/workercmd/cmd.go
```diff
@@ -70,6 +70,10 @@ func (m *Main) Run(ctx context.Context) error {
 		activities.NewMetadataValidationActivity().Execute,
 		temporalsdk_activity.RegisterOptions{Name: activities.MetadataValidationName},
 	)
+	w.RegisterActivityWithOptions(
+		activities.NewCombinePremisActivity().Execute,
```
djjuhasz, line 74: This looks like a good start @mcantelon. Just a quick note that Go initialisms like "PREMIS" should use the same case for all letters, see https://google.github.io/styleguide/go/decisions#initialisms. So `activities.NewCombinePremisActivity` should be `activities.NewCombinePREMISActivity`.

PR #26, opened 2024-06-07
Return a log of preprocessing activity outcomes

djjuhasz:
Return a log of preprocessing workflow activity outcomes to it's parent workflow.


internal/eventlog/eventlog.go
```diff
@@ -0,0 +1,47 @@
...
+	}
+}
+
+type Event struct {
+	Success     bool
+	Name        string
+	Message     string
+	StartedAt   time.Time
+	CompletedAt time.Time
+}
```
jraddaoui, line 24: Since this needs to be replicated in enduro, what are you expecting to find in the `Events[]` passed in the workflow result? My concern is that we are passing events only to create preservation tasks in enduro right now, so it may be better to call them `PreservationTaskEvents`, or just `PreservationTasks`. If you are envisioning this slice to contain other types of events, should it include a type field in the payload or find another way to differentiate even types in enduro?

I think that may be hard to maintain from both sides, so I'd rather call this slice `PreservationTasks` and have another one(s) in the future if we need to pass other types of events.

jraddaoui (changes requested):
Neat! I really like this implementation, thanks @djjuhasz!

Just a few minor comments below ;)

internal/workflow/preprocessing.go
```diff
@@ -20,14 +23,17 @@ type PreprocessingWorkflowParams struct {
 
 type PreprocessingWorkflowResult struct {
 	RelativePath string
+	EventLog     eventlog.Log
 }
 
 type PreprocessingWorkflow struct {
+	clock      clockwork.Clock
```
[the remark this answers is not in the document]
djjuhasz: Oops no, that's a left over from an earlier iteration.

internal/eventlog/eventlog.go
```diff
@@ -0,0 +1,47 @@
...
+	return e.complete(t, true, msg, a...)
+}
+
+func (e *Event) complete(t time.Time, success bool, msg string, a ...any) *Event {
+	e.CompletedAt = t
+	e.Success = success
+	e.Message = fmt.Sprintf(msg, a...)
+
+	return e
+}
```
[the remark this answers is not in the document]
djjuhasz: I prefer:
```
validateStructureEvent.fail(
    ctx,
    "Content error: SIP structure validation has failed:\n%s",
    strings.Join(validateStructure.Failures, "\n"),
)
```

to:

```
validateStructureEvent.fail(
    ctx, 
    fmt.Sprintf(
        "Content error: SIP structure validation has failed:\n%s",
        strings.Join(validateStructure.Failures, "\n"),
    ),
)
```

I'm lazy, so I don't like typing `fmt.Sprintf()` when I don't have to, especially for the `fail()` calls which all use string substitution. :)

internal/workflow/preprocessing.go
```diff
@@ -47,50 +55,109 @@ func (w *PreprocessingWorkflow) Execute(
...
-		return nil, e
+		identifySIPEvent.fail(
+			ctx,
+			"Content error: SIP structure could not be identified. Please ensure that SIP conforms to one of the following structures: %s",
+			strings.Join(enums.SIPTypeNames(), ", "),
+		)
+	} else {
+		identifySIPEvent.succeed(ctx, "SIP structure identified: %s", identifySIP.SIP.Type.String())
+	}
+	result.EventLog.Append(identifySIPEvent.Event)
```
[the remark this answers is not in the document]
djjuhasz: Right now the `IdentifySIP` activity can really only return a system error or an identified `sip.SIP`.  If I return a system error directly from the preprocessing workflow (i.e. `return nil, err`) then no event data will be returned to the parent workflow because Temporal workflows can not return both a result and an error (ref: https://community.temporal.io/t/failing-a-workflow-and-returning-data/6620/5).

So I think what I actually need to do here is:

1. If IdentifySIP returns an error, then record the error information in the identifySIPEvent and return that data in the `PreprocessingWorkflowResult` (e.g. `return PreprocessingWorkflowResult{eventLog: ...}, nil`)
2. If IdentifySIP doesn't return an error, update the identifySIPEvent appropriately, and continue the preprocessing workflow

internal/eventlog/eventlog.go
djjuhasz: Okay, to get into the pedantic semantics a bit here: _is_ SIP validation a preservation activity or is it a pre-ingest activity? :stuck_out_tongue_winking_eye: 

The `Event` struct doesn't map one-to-one to a [PreservationTask](https://github.com/artefactual-sdps/enduro/blob/main/internal/datatypes/preservation_task.go#L11), so I don't think it's necessary for them to use the same name.  I did intend for the `eventlog` package to be a fairly general event logging module, so it could be reused in other contexts (via copy/paste as it's an internal package).  I'm fine with adding a `Event.Type` field later if we need to distinguish an Event that will not become a PreservationTask, but that doesn't seem useful to me before we have distinct event types.

I think you raise a good point about defining the Event type here and in Enduro though - because Event is really defining an interface to Enduro, I think it's better to define the type in Enduro if possible, and then import it into preprocessing.  Alternatively we could create a protocol buffer definition of for Events (or PreservationTasks, or whatever we call it) and use that for communicating between Preprocessing and Enduro, though this seem like more work.

djjuhasz: I also just realized that these same events are going to be serialized as PREMIS events at some point, so maybe these events should be structured as PREMIS events?  PREMIS is pretty complex though and I don't want to complicate these events unless it's necessary. :thinking:

djjuhasz: > I think you raise a good point about defining the Event type here and in Enduro though - because Event is really defining an interface to Enduro, I think it's better to define the type in Enduro if possible, and then import it into preprocessing. Alternatively we could create a protocol buffer definition of for Events (or PreservationTasks, or whatever we call it) and use that for communicating between Preprocessing and Enduro, but this seem like more work.

Nevermind, importing the Event type from Enduro is probably a bad idea - the Event definition for this project should be owned by this project so external changes don't break the local implementation.

internal/workflow/event_wrapper.go
```diff
@@ -0,0 +1,23 @@
...
+	return &eventWrapper{eventlog.NewEvent(temporalsdk_workflow.Now(ctx), name)}
+}
+
+func (w *eventWrapper) fail(ctx temporalsdk_workflow.Context, msg string, a ...any) {
+	w.Fail(temporalsdk_workflow.Now(ctx), msg, a...)
+}
+
+func (w *eventWrapper) succeed(ctx temporalsdk_workflow.Context, msg string, a ...any) {
+	w.Succeed(temporalsdk_workflow.Now(ctx), msg, a...)
+}
```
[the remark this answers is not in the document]
jraddaoui: Yeah, I can see that. I still think the same, but not a big deal.

That reminds me, we should be using `workflowcheck` in CI to catch issues like that.

internal/eventlog/eventlog.go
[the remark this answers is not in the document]
jraddaoui: I can see that too, I just prefer the later, the first one feels like a bit of magic. And I don't know how expensive is to pack/unpack those parameters, I guess not a lot, so not a big deal either.

internal/workflow/preprocessing.go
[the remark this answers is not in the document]
jraddaoui: I read the conversation you had with Sevein in Slack. I was just thinking on changing that activity as you did with the others to follow the same pattern in workflow, at some point it may return a system error instead of an identification error. Also, not a big deal.

internal/eventlog/eventlog.go
jraddaoui: Don't worry too much about the naming or if they match one to one. My take is that an event type is hard to maintain across multiple repositories, and that re-using the same payload for different events may be more prone to backward incompatible changes.

internal/workflow/preprocessing.go
```diff
@@ -20,14 +23,17 @@ type PreprocessingWorkflowParams struct {
 
 type PreprocessingWorkflowResult struct {
 	RelativePath string
+	EventLog     eventlog.Log
 }
```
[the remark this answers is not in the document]
jraddaoui: We can follow-up in the other thread if needed.

internal/workflow/event_wrapper.go
```diff
@@ -0,0 +1,26 @@
...
+	return &eventWrapper{eventlog.NewEvent(temporalsdk_workflow.Now(ctx), name)}
+}
+
+func (w *eventWrapper) Complete(
+	ctx temporalsdk_workflow.Context,
+	outcome enums.EventOutcome,
+	msg string,
+	a ...any,
+) *eventWrapper {
+	w.Event.Complete(temporalsdk_workflow.Now(ctx), outcome, msg, a...)
```
jraddaoui, line 24: If we still want to keep the wrapper, can we format the message here to avoid packing and unpacking again?

internal/workflow/preprocessing.go
```diff
@@ -1,25 +1,44 @@
...
 )
 
 type PreprocessingWorkflowParams struct {
 	RelativePath string
 }
 
 type PreprocessingWorkflowResult struct {
+	Outcome      Outcome
 	RelativePath string
+	Events       []eventlog.Event
```
jraddaoui, line 35: Feel free to ignore, but I'll insist. I think `Events` is too general if they are only meant to create preservation tasks in Enduro, I'd call them `PreservationTasks` directly.

internal/workflow/preprocessing.go
```diff
@@ -36,94 +55,225 @@ func (w *PreprocessingWorkflow) Execute(
...
 	// against the checksums generated on Bag creation.
 
-	return &PreprocessingWorkflowResult{RelativePath: params.RelativePath}, e
+	return &result, nil
 }
 
 func withLocalActOpts(ctx temporalsdk_workflow.Context) temporalsdk_workflow.Context {
 	return temporalsdk_workflow.WithActivityOptions(ctx, temporalsdk_workflow.ActivityOptions{
-		ScheduleToCloseTimeout: 5 * time.Minute,
+		ScheduleToCloseTimeout: time.Minute,
```
jraddaoui, line 256: Why this change? Testing leftover? I'd keep it as 5 minutes

internal/workflow/preprocessing.go
```diff
@@ -36,94 +55,225 @@ func (w *PreprocessingWorkflow) Execute(
...
+func systemError(
+	logger temporalsdk_log.Logger,
+	activityName string,
+	result *PreprocessingWorkflowResult,
+	err error,
+) (*PreprocessingWorkflowResult, error) {
+	logger.Error(activityName, "system error", err.Error())
+	result.Outcome = OutcomeSystemError
+	return result, nil
+}
```
jraddaoui, line 279: Not a big deal either, feel free to ignore.

These two functions feel like a bit of magic to me, with `nolint:unparam` and returning things that should be needed just to be able to use them directly from a return call. I get why you create them, but I think the first one is not needed (only used once), and the second one I'd have it without return, just logging and updating the result, then calling return after the function call.

internal/workflow/preprocessing.go
```diff
@@ -1,25 +1,44 @@
...
+	"github.com/artefactual-sdps/preprocessing-sfa/internal/eventlog"
+)
+
+type Outcome int
+
+const (
+	OutcomeSuccess Outcome = iota
+	OutcomeSystemError
+	OutcomeContentError
 )
```
jraddaoui, line 26: I like the outcome being an `int`, it should be easier to document and match in enduro.

internal/eventlog/eventlog.go
```diff
@@ -0,0 +1,35 @@
+package eventlog
```
jraddaoui, line 1: We may want to rename this package to just `event`.

internal/enums/event_outcome.go
```diff
@@ -0,0 +1,10 @@
+package enums
+
+// ENUM(
+// Unknown,
+// Success,
+// Failure,
+// Valid,
+// Invalid,
+// ).
+type EventOutcome string
```
jraddaoui, line 10: I feel like `Success` and `Valid` overlap with each other and that `Unknown` is not used. I prefer something similar to the workflow outcome values, for example: `Success`, `ValidationFailure` and `SystemFailure`.

jraddaoui (changes requested):
Thanks @djjuhasz! Again, what I think are minor comments ;) But it's looking great.

internal/workflow/event_wrapper.go
djjuhasz: Sure.

internal/workflow/preprocessing.go
djjuhasz: I changed it because I can't imagine any of the preprocessing activities taking more than a minute to complete.  I'll change it back to 5 minutes though, if you think a minute is too low.

internal/eventlog/eventlog.go
djjuhasz: I choose the name based on https://brandur.org/fragments/go-no-common-nouns.

internal/enums/event_outcome.go
djjuhasz: The "Unknown" value was inspired by Enduro, which often has a first enum value of "Unspecified" (e.g. https://github.com/artefactual-sdps/enduro/blob/main/internal/enums/preservation_task.go). I chose "Unknown" based on a faulty memory of the Enduro `enums.PreservationTask` value, I think "Unspecified" is better.  

I think the default value should not assume an actual outcome like "Success" or "Failure", because when an event instance is created the event is not complete, so it has no outcome.  I will change the Event constructor to set the default value of `Event.Outcome` to `Unspecified`.

I'll change the other values to Success, ValidationFailure and SystemFailure, as you've suggested.

internal/workflow/preprocessing.go
djjuhasz: I'll get rid of the unused "error" return, which removes the need for the `nolint:unparam` directive.

Even though `contentError()` is only called once,  I think it's useful as documentation in code of what the expected returns from a content error are - a developer can easily look at the `contentError()` and `systemError()` function definitions to understand what the workflow should return in each case.

internal/eventlog/eventlog.go
```diff
@@ -18,12 +18,13 @@ type Event struct {
...
 	}
 }
 
-func (e *Event) IsValid() bool {
-	return e.Outcome == enums.EventOutcomeValid
+func (e *Event) IsSuccess() bool {
+	return e.Outcome == enums.EventOutcomeSuccess
 }
 
 func (e *Event) Complete(t time.Time, outcome enums.EventOutcome, msg string, a ...any) *Event {
```
jraddaoui, line 30: `a` param is no longer needed after the changes in the wrapper, and `msg` should be already formatted and can be used directly to set the message value.

jraddaoui (approved):
Thanks @djjuhasz!

internal/eventlog/eventlog.go
djjuhasz: Huh, interesting. I don't have the same aversion that you do to using `any` in a typed language, but I can understand why you dislike it.  I do think `fmt.Sprintf()` is a good example of a function that demonstrates why allowing an `any` parameter type is necessary in a typed GPL.

I did notice while writing the code that the hints from `fmt.Sprintf()` bubble up to the wrapper functions, which _is_ a bit magical, but I think it's magical in a good way. :)

PR #31, opened 2024-07-05
Fixes to PREMIS event recording

mcantelon:
* Removed redundant PREMIS event for validate file formats
* Correct use of improper PREMIS event types
* Allow PREMIS evebt details and PREMIS event outcome detail to be specified rather than generated
* Add PREMIS event outcome detail specification to PREMIS events in workflow and validate file formats activity
* Generate error when attempting to add a PREMIS event to a non-existent PREMIS object
* Fixed issue with validate formats activity
* Removed unneeded code in tests for add PREMIS event activity

### DIFF
--- a/internal/premis/premis.go
+++ b/internal/premis/premis.go
@@ -32,9 +32,10 @@ type Object struct {
 }
 
 type EventSummary struct {
-	Type    string
-	Detail  string
-	Outcome string
+	Type          string
+	Detail        string
+	Outcome       string
+	OutcomeDetail string
 }
 
 type Event struct {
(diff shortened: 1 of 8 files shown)


internal/premis/premis.go
```diff
@@ -128,30 +129,23 @@ func eventFromEventSummaryAndAgent(eventSummary EventSummary, agent Agent) Event
 	}
 }
 
-func NewEventSummary(eventType string, failures []string) (EventSummary, error) {
-	detail, outcome := assembleEventDetailAndOutcome(failures)
+func NewEventSummary(eventType string, detail string, outcomeDetail string, failures []string) (EventSummary, error) {
```
jraddaoui, line 132: Do we still need to pass the slice of failures, could we use a boolean or pass the outcome directly.

jraddaoui (approved):
LGTM!

internal/premis/premis.go
mcantelon: Good call... that'd make it more straightforward.

mcantelon: (Fixed.)

mcantelon:
Thanks @jraddaoui !

PR #36, opened 2024-08-01
Add "born digital" SIP type

### LINKED ISSUE #34: Feature: Add support for Born digital SIPs and update SIP identification rules

SFA has decided to refine the identification of SIP types, in a way that better reflects the business logic that will be applied to the SIP as it is processed. Instead of Vecteur SIP, GEVER SIP, SIARD SIP, etc., there should be three branches for identification:

* **Born digital SIP**: any eCH-0160 compliant SIP
* **Digitized SIP**: what we're currently calling a Vecteur SIP
* **Digitized AIP**: what we're currently calling a Vecteur AIP

This should be pretty straightforward. At this time, we are not being asked to add any validation logic for Digital born SIPs beyond the XSD checks already in place - we will just presume that any transfer that makes it to DPS conforms to eCH-0160, so we can treat this as the basic type.

Christa has suggested the following differentiators for Digitized SIPs:

* Vecteur SIPs always have the word “Vecteur” in the SIP name
* Vecteur SIPs have PREMIS metadata, one per package (`Prozess_Digitalisierung_PREMIS.xml`) and one for each file in the package (e.g. `00000001.jp2` will have a corresponding `00000001_PREMIS.xml`)

For Digitized AIPs, we can use the same identification check as is currently in place - looking for the `additional` directory.

Just as a note Digitized AIPs will also have the same PREMIS metadata as Digitized SIPs; however, they do not always have the word "Vecteur" in the AIP name.

**Additional context**

djjuhasz:
Fixes #34.

- Add "born digital" SIP type and add it to the "identify SIP" activity,
- Replace "Vecteur" with "digitized" for SIP types

### DIFF
--- a/internal/activities/identify_sip.go
+++ b/internal/activities/identify_sip.go
@@ -28,5 +28,5 @@ func (a *IdentifySIP) Execute(ctx context.Context, params *IdentifySIPParams) (*
 		return nil, fmt.Errorf("IdentifySIP: %v", err)
 	}
 
-	return &IdentifySIPResult{SIP: s}, nil
+	return &IdentifySIPResult{SIP: *s}, nil
 }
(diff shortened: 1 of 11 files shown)


jraddaoui (approved):
Nice, thanks @djjuhasz!

Issue #22, opened 2024-05-31
Feature: Improve validation scripts to use Go and package XSDs


sallain:
**Is your feature request related to a problem? Please describe.**

Pilot project work made use of Python scripts provided by SFA. In PoC#1 we did a bit of cleanup, adding some golang wrappers and handling. However, as SFA's needs evolve, the original scripts need some updating. Rather than updating the Python and having to maintain that dependency, we should rewrite them in Go so that they are aligned with Enduro's core language.

A second aspect of the validation scripts that we should amend is the use of locally stored XSDs. This was acceptable when we were only validating one type of SIP, but now we must support multiple transfer types which may comply with a different version of the XSDs. We should use the XSDs that are included in each SIP in order to ensure that the correct XSD version is used.

**Describe the solution you'd like**

This is a catch-all card for updating the scripts in general. The following list is a start:

- [ ] Rewrite the overall validation to use golang
- [ ] Use our existing tools for bagging and bag validation
- [ ] Use whatever XSDs are included in the transfer for validation, instead of a local copy
- [ ] Generalize the implementation so that task outcomes and details can be shown in the Enduro UI, per [this separate dev card for PoC#2](https://miro.com/app/board/uXjVKWYxPh4=/?openCardPanel=3458764589904104783)

More tasks can be added to this card as needed!

**Describe alternatives you've considered**

None

**Additional context**

fiver-watson:
@jraddaoui I would be okay with doing it that way. However, another factor for consideration... (cc @sallain  to make sure i am getting this right): 

During our most recent metadata WG meeting, SFA shared what they would ideally like to see us do with the metadata.xml / UpdatedArelda file - i.e. extract the userful information and turn it into other AM inputs to be included in the SIP, as well as create a version to send to the AIS (details on that to come - waiting on a copy of their current XSD and some example input  / output files). See slide 2 in this deck from SFA: 

* https://docs.google.com/presentation/d/1kkxvjLJ8OwdLo8FujHJQRNZ5eGOgOC6l/edit#slide=id.p2

[...]

PR #38, opened 2024-08-02
Add "verify manifest" activity, fixes #35

### LINKED ISSUE #35: Feature: ensure that each file listed in metadata.xml actually exists in the SIP

The metadata.xml file (for Digital born SIPs and Digitized SIPs) or UpdatedAreldaMetadata.xml file (for Digitized AIPs) contains a list of all objects in the content directory of the SIP. It is possible that the metadata file and the actual objects in the SIP might not match up. We should confirm that all of the objects listed in the metadata exist in the SIP, and vice versa.

There are a few different ways to complete this check. One option is to use the checksums provided in the metadata file. In a separate issue (#22), I proposed that we should validate the incoming checksums from metadata.xml or UpdatedAreldaMetadata.xml before bag creation. We could extend this to accomplish two validations: ensuring that all of the items in the metadata file exist, and ensuring that they are uncorrupted.

This check should go both ways:
* If there is a checksum but no object, this should result in an error.
* If there is an object that does not have a corresponding checksum, this should also result in an error.

If we wanted to separate the checksum check from the object check, we could just use the file listing within metadata.xml or UpdatedAreldaMetadata.xml. Each file in the SIP is listed in `<ordner><content><datei>`, with the full name including extension in the `<name>` element.

```
<ordner>
    <name>content</name>
    <originalName>content</originalName>
    <ordner>
        <name>d_0000001</name>
        <originalName>d_0000001</originalName>
        <datei id="_zodSTSD0nv05CpOp6JoV3X">
            <name>00000001.jp2</name>
            <originalName>00000001.jp2</originalName>
            <pruefalgorithmus>MD5</pruefalgorithmus>
            <pruefsumme>dc29291d0e2a18363d0efd2ec2fe81c9</pruefsumme>
        </datei>
        <datei id="_rlPKJX9ZcAl4ooc4IfoIkM">
            <name>00000002.jp2</name>
            <originalName>00000002.jp2</originalName>
            <pruefalgorithmus>MD5</pruefalgorithmus>
            <pruefsumme>9093907ec32f06fe595e0f14982c4bf0</pruefsumme>
        </datei>
    </ordner>
</ordner>
```

**Additional context**

djjuhasz:


internal/activities/verify_manifest_test.go
```diff
@@ -0,0 +1,136 @@
...
+		filepath.Join(extraFile, "content", "content", "d_0000001", "extra_file.txt"),
+		[]byte("I'm an extra file."),
+		0o600,
+	)
+	if err != nil {
+		t.Fatalf("setup: write file: %v", err)
+	}
+
+	return td
+}
```
jraddaoui, line 52: It would be nice to fake the SIP structure and XML contents instead of having one in that testdata folder, it shouldn't be hard to do, there are some examples across the tests already:

https://github.com/artefactual-sdps/preprocessing-sfa/blob/main/internal/activities/transform_sip_test.go

I doesn't need to be a full SIP with the XSD files and all to prove the activity behavior.

jraddaoui (approved):
Nice! Feel free to ignore the comment below.

internal/activities/verify_manifest_test.go
djjuhasz: Yeah, good idea @jraddaoui.  I was originally going to check the file checksums too, which would require actual file contents to checksum, but after rereading the comments on #22 I noticed that we may be checking the metadata.xml checksums against the BagIt bag checksums after we generate the bag.  In any case, I think it's better to check the checksums in a separate activity.

TIL; I'll remove the test SIP and fake the XML + structure in the test file.

djjuhasz:
@jraddaoui after removing the test SIP and rebasing to get the latest updates from main, I've made quite a few change to the original PR.  Can you take another look at the PR please?

internal/sip/sip.go
```diff
@@ -11,29 +11,49 @@ import (
...
+	MetadataPath string
+
+	// UpdatedAreldaMDPath is the filepath of the UpdatedAreldaMetadata.xml file
+	// for digitized AIPs — it is empty for all other SIP types.
 	UpdatedAreldaMDPath string
-	XSDPath             string
-	TopLevelPaths       []string
+
+	// XSDPath is the filepath of the "xsd" directory.
+	XSDPath string
```
jraddaoui, line 36: It's the path of the "arelda.xsd" file.

internal/sip/sip.go
```diff
@@ -11,29 +11,49 @@ import (
...
 
-func NewSIP(path string) (*SIP, error) {
-	s := SIP{Path: path}
+func New(path string) (SIP, error) {
+	s := SIP{}
 
-	if _, err := os.Stat(s.Path); err != nil {
-		return nil, fmt.Errorf("NewSIP: %v", err)
+	if _, err := os.Stat(path); err != nil {
+		return s, fmt.Errorf("NewSIP: %v", err)
```
jraddaoui, line 46: We should tweak the error messages too, no longer called `NewSIP`.

jraddaoui (approved):
LGTM, just a couple nitpicks, thanks @djjuhasz!

jraddaoui (changes requested):
One thing I missed from https://github.com/artefactual-sdps/preprocessing-sfa/commit/6bd563e6f3cd237511b03ad2e9ddde05fe18c67a.

I think we should be validating `UpdatedAreldaMetadata.xml` for digitized AIPs and, with the changes in the commit above, we always validate `metadata.xml`. Now that you added  the `ManifestPath`, we should use that one calling the validation activity, here:

https://github.com/artefactual-sdps/preprocessing-sfa/blob/main/internal/workflow/preprocessing.go#L215

internal/sip/sip.go
djjuhasz: Ah, okay — I'll update the comment. :)

djjuhasz:
@jraddaoui I added a commit to fix the digitized AIP metadata validation — I think I'll keep it as a separate commit when I merge to main.

I also found another bug where the processing workflow wasn't halted when "verify manifest" failed.  I've added a third commit to fix this bug, which I'll squash into the first commit before merging.

jraddaoui:
Good catch, thanks @djjuhasz! I've done the squash and I'll merge to be able to deploy these changes.

jraddaoui (approved):

PR #42, opened 2024-08-13
Add AIS integration PoC

jraddaoui:
Create a new temporal client, worker and workflow to fetch AIP
metadata from Archivematica's Storage Service after AIPs are stored,
add it to a zipped folder and upload it to a S3/MinIO bucket.

Add a small API server to receive post-store AIP service callbacks
from Archivematica's Storage Service and start the AIS workflow.

Run the workers and the API in the same command for now, using
github.com/oklog/run to manage goroutines.


cmd/worker/main.go
```diff
@@ -65,23 +71,90 @@ func main() {
...
+	tc, err := temporalsdk_client.Dial(temporalsdk_client.Options{
+		HostPort:  cfg.Temporal.Address,
+		Namespace: cfg.Temporal.Namespace,
+		Logger:    temporal.Logger(logger.WithName("temporal")),
+	})
+	if err != nil {
+		logger.Error(err, "Unable to create Temporal client.")
+		os.Exit(1)
+	}
+	defer tc.Close()
```
jraddaoui, line 84: Moving the Temporal client creation outside `cmd.go`, to be able to share it with the API server.

cmd/worker/main.go
```diff
@@ -65,23 +71,90 @@ func main() {
...
+			},
+		)
+	}
+
+	err = g.Run()
+	if err != nil {
+		logger.Error(err, "Application failure.")
 		os.Exit(1)
 	}
+	logger.Info("Bye!")
```
jraddaoui, line 159: I just copied this `run.Group` implementation from Enduro to be able to run the worker and API server in the same process, there are some things I don't fully understand, but I hope we can remove the API soon and trigger this workflow from Enduro directly.

cmd/worker/workercmd/cmd.go
```diff
@@ -2,8 +2,9 @@ package workercmd
 
 import (
 	"context"
+	"fmt"
 
-	"github.com/artefactual-sdps/temporal-activities/bagit"
+	"github.com/artefactual-sdps/temporal-activities/bagcreate"
```
jraddaoui, line 7: Updating temporal-activities after the re-org. I'll do it in other projects too ;)

cmd/worker/workercmd/cmd.go
```diff
@@ -96,10 +89,15 @@ func (m *Main) Run(ctx context.Context) error {
...
-		temporalsdk_activity.RegisterOptions{Name: bagit.CreateBagActivityName},
+		bagcreate.New(m.cfg.Bagit).Execute,
+		temporalsdk_activity.RegisterOptions{Name: bagcreate.Name},
 	)
 
+	// AIS workflow.
+	if err := ais.RegisterWorkflow(ctx, w, m.cfg.AIS); err != nil {
+		return fmt.Errorf("AIS: %w", err)
+	}
+
```
jraddaoui, line 100: I decided to do it this way to keep AIS integration things together as much as possible, to ease its extraction to a different worker if we need to.

internal/ais/amss.go
```diff
@@ -0,0 +1,95 @@
...
+		return "", fmt.Errorf("GetAIPPath: current_path not found in response")
+	}
+
+	path, ok := value.(string)
+	if !ok {
+		return "", fmt.Errorf("GetAIPPath: current_path is not a string")
+	}
+
+	return path, nil
+}
```
jraddaoui, line 66: I thought we would need this method to get the relative path from the AMSS after reading [the Wiki](https://wiki.archivematica.org/Storage_Service_API#Download_single_file). However, we have the AIP name and the AIP UUID already, so we can build it without hitting this endpoint. I'll double check with the maintainers to see if that's always true and remove this method if it is.

internal/ais/amss.go
```diff
@@ -0,0 +1,95 @@
...
+func (c *AMSSClient) DownloadAIPFile(ctx context.Context, aipUUID, path string, writer io.Writer) error {
+	u := c.url.JoinPath("api/v2/file", aipUUID, "extract_file")
+	query := url.Values{}
+	query.Set("relative_path_to_file", path)
+	u.RawQuery = query.Encode()
+	req, err := http.NewRequestWithContext(ctx, "GET", u.String(), nil)
+	if err != nil {
+		return fmt.Errorf("DownloadAIPFile: new request: %w", err)
+	}
+	req.Header.Set("Authorization", c.auth)
```
jraddaoui, line 77: I thought about using a custom transport to do this as you did in Enduro, but if we end up only with this method it seems unnecessary.

internal/ais/api.go
```diff
@@ -0,0 +1,61 @@
...
+			AIPUUID: aipUUID,
+			AIPName: payload.Name,
+		})
+		if err != nil {
+			http.Error(w, "Couldn't start AIS workflow", http.StatusInternalServerError)
+			return
+		}
+
+		w.WriteHeader(http.StatusAccepted)
+	})
```
jraddaoui, line 52: Service callback configuration needed in the AMSS to hit this endpoint, running all locally:

- Event: `Post-store AIP`
- URI: `http://host.docker.internal:9035/stored`
- Method: `POST`
- Headers (key/value): `Content-type` `application/json`
- Body:
  ```
  {
    "uuid": "<package_uuid>",
    "name": "<package_name>"
  }
  ```
- Expected Status: `202`

I'll document this properly if we decide to use this approach in the end.

If you try this locally using AM docker env, you will need to add the following to the `archivematica-storage-service` container in the `docker-compose` file:

```
extra_hosts:
  - "host.docker.internal:host-gateway"
```

internal/ais/parse.go
```diff
@@ -0,0 +1,128 @@
...
+			return nil, fmt.Errorf("ParseActivity: parse METS: %w", err)
+		}
+
+		switch elem := tok.(type) {
+		case xml.StartElement:
+			switch elem.Name.Local {
+			case "fileGrp":
+				if getAttr(elem, "USE") == "metadata" {
+					inFileGrp = true
+				}
```
jraddaoui, line 63: I'll need to change this after what we discussed today to be able to find the `metadata.xml` outside the metadata folder.

internal/ais/parse.go
```diff
@@ -0,0 +1,128 @@
...
+					currentFileID = getAttr(elem, "ID")
+				}
+			case "FLocat":
+				if currentFileID != "" {
+					mdFiles[currentFileID] = getAttr(elem, "href")
+				}
+			case "structMap":
+				if getAttr(elem, "TYPE") == "physical" {
+					inStructMap = true
+				}
```
jraddaoui, line 75: Also add a `inHeadDir` or something like that with an extra check to be "sure" we are not picking another metadata.xml file.

jraddaoui (commented):
This is ready for an initial review @djjuhasz.

internal/ais/workflow.go
```diff
@@ -0,0 +1,222 @@
...
+
+type Workflow struct {
+	workingDir string
+}
+
+func NewWorkflow(workingDir string) *Workflow {
+	return &Workflow{workingDir: workingDir}
+}
+
+func (w *Workflow) Execute(ctx temporalsdk_workflow.Context, params *WorkflowParams) (r *WorkflowResult, e error) {
```
jraddaoui, line 42: I'll make sure all the activities from this workflow are executed under a session. Right now, because we run the preprocessing child workflow in the same worker, which requires to share the filesystem with Enduro, it won't be an issue if the activities are run from different workers. But the same requirement doesn't apply to this workflow so we could create another worker later on and scale it differently.

internal/ais/amss.go
```diff
@@ -0,0 +1,95 @@
...
+	"encoding/json"
+	"fmt"
+	"io"
+	"net/http"
+	"net/url"
+
+	"github.com/hashicorp/go-cleanhttp"
+)
+
+type AMSSClient struct {
```
djjuhasz, line 14: Why didn't you use https://github.com/artefactual-labs/ssclient-go?

jraddaoui: I took a look and I think it only covers a few endpoints from the `location` and `pipeline` namespaces, nothing from the `file` namespace. I thought you didn't use it in Enduro for the same reason.

internal/ais/workflow.go
```diff
@@ -0,0 +1,222 @@
...
+	aipDirName := fmt.Sprintf("%s-%s", params.AIPName, params.AIPUUID.String())
+	localDir := filepath.Join(w.workingDir, fmt.Sprintf("search-md_%s", aipDirName))
+	metsName := fmt.Sprintf("METS.%s.xml", params.AIPUUID.String())
+	metsPath := filepath.Join(localDir, metsName)
+
+	removePaths = append(removePaths, localDir)
+
+	var fetchMETSResult FetchActivityResult
+	e = temporalsdk_workflow.ExecuteActivity(
+		withRemoteActOpts(ctx),
```
djjuhasz, line 72: I don't think the fetch activity is sending a heartbeat, so it will timeout if the download time exceeds the heartbeat timeout of one minute.

djjuhasz (commented):
Wow, this looks really good overall @jraddaoui! :open_mouth:  

I noted a few potential problems, and it seems pretty light on unit tests right now.  I'm assuming as this is a WIP the tests are still to come. :)

internal/ais/parse.go
```diff
@@ -0,0 +1,128 @@
...
+)
+
+const ParseActivityName = "parse-mets-file"
+
+type (
+	ParseActivityParams struct {
+		METSPath string
+	}
+	ParseActivityResult struct {
+		MdRelPath   string
```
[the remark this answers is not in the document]
djjuhasz: Oh I see: "UpdatedAreldaMetadataRelativePath".  I know it's a mouthful (or maybe a "handful" to type?) but I'd be tempted to use the full, unabbreviated name for clarity.  If you don't want a very long field name then comments would be really helpful. :wink:

internal/ais/workflow.go
jraddaoui: Good catch, thanks!

internal/ais/parse.go
```diff
@@ -0,0 +1,128 @@
...
+					if label == "metadata.xml" {
+						inMdDiv = true
+					}
+				}
+			case "fptr":
+				if inUAMdDiv {
+					uamdFileID = getAttr(elem, "FILEID")
+				}
+				if inMdDiv {
+					uamdFileID = getAttr(elem, "FILEID")
```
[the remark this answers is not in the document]
jraddaoui: :facepalm: thanks!

internal/ais/parse.go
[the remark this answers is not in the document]
jraddaoui: I'll use the long name and add comments too ;)

internal/ais/amss.go
djjuhasz: Ah! I honestly don't remember creating an AMSS API client in Enduro, are you sure it was me?  I can't even find an AMSS client in the Enduro code, but maybe I'm looking in the wrong place. :confused:

jraddaoui:
Thanks @djjuhasz, I'll address those points before merging. Tests will probably come in a different PR, I want to check about possible follow-ups with the team before getting too tied with the implementation, what I mentioned in the issue.

internal/ais/amss.go
jraddaoui: Maybe it wasn't you. It's actually called `ssblob`, looks to be an abstraction for the AMSS location type:

https://github.com/artefactual-sdps/enduro/tree/main/internal/storage/ssblob

djjuhasz: Oh yeah, that was Keith and Jesús. :) I think the ssblob work predates https://github.com/artefactual-labs/ssclient-go.  It doesn't really matter if ssclient-go doesn't work with the endpoints you need.

Issue #19, opened 2024-05-30
Feature: capture PREMIS events for pre-ingest validation tasks


sallain:
**Is your feature request related to a problem? Please describe.**

SFA SIPs will have some custom ingest validation tasks in their workflow that will: 

- Validate the transfer structure
- Validate the metadata files included in the transfer
- Check the file formats included against a controlled list

At present, none of these pre-ingest validation tasks are generating PREMIS events. This card aims to change that where possible.

In some cases it would be best to create PREMIS events at the package level (such as for the transfer structure validation) - since AM can't do right now, we will focus only on those events we can add at the file level. For now, we will focus on generating a validation event for each file in a package once it has been checked against the allowed file formats list during the ingest validation phase. 

**Describe the solution you'd like**

Generate file-level PREMIS events where possible, and include them as part of a new  well-formed premis.xml file.

The first candidate is file format validation.

- Generate one PREMIS event for each object as it is checked against the allowed list. Sample PREMIS event [here](https://miro.com/app/board/uXjVMlYiVgs=/?moveToWidget=3458764588457606174&cot=14)
- Capture validation events in a single premis.xml file (see comments below for an annotated PREMIS file that you can use as a template)
- Place the newly-generated premis.xml file in the metadata directory

**Describe alternatives you've considered**

**Additional context**

Previously, we were generating a premis.xml file by combining individual PREMIS files found within the content directory. This work is being undone by #18.

sallain:
Annotated PREMIS file for multiple objects: 
[premis-annotated-multi.zip](https://github.com/artefactual-sdps/preprocessing-sfa/files/15506203/premis-annotated-multi.zip)

sallain:
In the sample PREMIS event, @fiver-watson provided a generic linkingAgentIdentifierValue of `<premis:linkingAgentIdentifierValue>https://github.com/artefactual-sdps/preprocessing-base</premis:linkingAgentIdentifierValue>` on the principle that this kind of validation is likely to be universally useful. However, since it's currently implemented for SFA only, I think I would recommend pointing to this repo as the Agent (that is, the child workflow is the agent).

sallain:
The structure of the PREMIS file looks good. However, I'm seeing two issues:

1. Archivematica isn't able to load the events from the premis.xml due to incorrect associations between events and objects.

[...]

PR #44, opened 2024-08-18
Fixes/updates to PREMIS event recording (#19)

mcantelon:
* Renamed event adding functions
* Fixed PREMIS event recording to add one event per file
* Updated PREMIS event recording and tests to work with updated transfer structure


jraddaoui (approved):
Thanks @mcantelon!

PR #45, opened 2024-08-21
WIP: integrate xml validator

mcantelon:


mcantelon:
Will add this to Enduro workflow.

PR #46, opened 2024-08-22
Verify the SIP header files match the manifest

djjuhasz:
Refs #35

- Add the SIP "header" directory to the files checked against the manifest
- Add an exception to ignore the "header/metadata.xml" file when it's not included in the manifest (digitized SIP, born digital)
- Add a unit test that verifies a digitized SIP manifest


jraddaoui (approved):
Thanks @djjuhasz!

PR #49, opened 2024-08-28
Fix born digital SIP PREMIS event recording (#19)

mcantelon:
Added support for born digital SIPs to PREMIS event recording.


jraddaoui (approved):
Thanks @mcantelon!

PR #51, opened 2024-08-28
Tweak DigitizedSIP identification

jraddaoui:
Use case-insensitive match to find "vecteur" in the SIP path.

Refs #34.


mcantelon (approved):
LGTM!

PR #52, opened 2024-08-28
Fix PREMIS event recording for all SIP types (#19)

mcantelon:


internal/premis/premis.go
```diff
@@ -422,13 +422,13 @@ func FilesWithinDirectory(contentPath string) ([]string, error) {
...
+	// Handle AIP structure differently from SIP structure.
 	var transferDirName string
 
-	if sip.Type == enums.SIPTypeBornDigital {
-		transferDirName = filepath.Base(filepath.Dir(sip.ContentPath))
-	} else {
+	if sip.Type == enums.SIPTypeDigitizedAIP {
 		transferDirName = filepath.Base(filepath.Dir(filepath.Dir(sip.ContentPath)))
+	} else {
+		transferDirName = filepath.Base(filepath.Dir(sip.ContentPath))
```
jraddaoui, line 431: I think we could use `filepath.Base(sip.Path)`. That would work for all types, right? No need for the if and so on.

PR #54, opened 2024-09-06
Verify manifest checksums match file contents

### LINKED ISSUE #40: Feature: validate file checksums from XML metadata

The initial Python script used for Bag creation ([here](https://github.com/artefactual-sdps/enduro-sfa/blob/stable/hack/sampledata/xsd/repackage_sip.py)) was creating the Bag manually, using the checksums from the XML metadata file to create the manifest, and then validating the final Bag with [bagit-python](https://github.com/LibraryOfCongress/bagit-python).

We changed the Bag creation process to use [go-bagit](https://github.com/nyudlts/go-bagit), that creates the manifest automatically so we are loosing that validation.

We are adding an activity to check the existence of the files from XML metadata in the SIP ([here](https://github.com/artefactual-sdps/preprocessing-sfa/pull/38)), but that is not currently validating those checksums either.

Add checksum validation in a new activity or include it in the activity mentioned above.

**Additional context**

See the discussion about where should we place that validation, we may want to do it alongside all the validation tasks (generating new checksums) or after Bag creation (re-using the checksums generated in that process):

https://github.com/artefactual-sdps/preprocessing-sfa/issues/22#issuecomment-2145339181

djjuhasz:
Fixes #40.

- Add manifest package with a Files() function that stream parses an SFA
  XML manifest and returns a map of filepaths to checksums
- Switch to `manifest.Files()` to parse SFA manifest files more
  efficiently
- Add checksum verification to the "verify manifest" activity
- Move "internal/workflow/testdata/little-Test-AIP-Digitization" to
  "internal/testdata/little-Test-AIP-Digitization"
- Update "tesdata/little-Test-AIP-Digitization" file hashes to match
  file contents
- Add a verify checksums event to the preprocessing workflow


djjuhasz:
Screenshot of the checksum verification failing:
Screenshot from 2024-09-06 11-13-48

mcantelon (approved):
Other than one possible typos, all looks good to me! Very nice!

PR #56, opened 2024-09-13
Use manifest file id as PREMIS objectIdentifier

### LINKED ISSUE #53: Feature: import file identifiers from metadata.xml into METS

Each file in an SFA SIP has a unique file identifier, which is listed in the package's metadata.xml file as `<datei id="_zodSTSD0nv05CpOp6JoV3X">`. This identifier is widely used in AIS to connect information across many sources. SFA would like this identifier to be present in the METS file so that the METS data can also be connected using the identifier.

There are a few different options we could implement:

1. We could add a second `<premis:objectIdentifier>` to the premis.xml implementation that is already in place; alternately, I _think_ that the premis.xml relies on `<premis:originalName>` to connect the premis object to the METS metadata, so we could perhaps have JUST the SFA identifier as the `<premis:objectIdentifier>`.
2. We could use Archivematica's [identifiers.json](https://www.archivematica.org/en/docs/archivematica-1.16/user-manual/transfer/transfer/#transfer-pids) feature to bypass doing the above PREMIS work, which has the same result (creates a second `<premis:objectIdentifier>`)

These are the two options I could think of!

**Additional context**

djjuhasz:
Fixes #53.

- Parse manifest datei id attributes
- Use the datei id value as the PREMIS objectIdentifierValue

### DIFF
--- a/internal/activities/add_premis_objects.go
+++ b/internal/activities/add_premis_objects.go
@@ -2,11 +2,11 @@ package activities
 
 import (
 	"context"
+	"fmt"
 	"os"
 	"path/filepath"
 
-	"github.com/google/uuid"
-
+	"github.com/artefactual-sdps/preprocessing-sfa/internal/manifest"
 	"github.com/artefactual-sdps/preprocessing-sfa/internal/premis"
 	"github.com/artefactual-sdps/preprocessing-sfa/internal/sip"
 )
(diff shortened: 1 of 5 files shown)


djjuhasz:
Unfortunately the `<objectIdentifierValue>` in the premis.xml is not preserved in the Archivematica METS file (see https://github.com/artefactual-sdps/preprocessing-sfa/issues/53#issuecomment-2350233801) so this PR doesn't actually fix #53. :(

PR #61, opened 2024-10-08
Read allowed file formats from a CSV file

### LINKED ISSUE #60: Feature: make allow file format lists configurable

The allow file format list is currently hardcoded. Anyone who wants to use this feature will have their own list of allowed formats, so they would have to change the code to suit their requirements. This also puts the list at risk during upgrades/redeployments as it may be reset to the hardcoded default.

Make the allow file format list configurable.

In a short discussion with @djjuhasz, he suggested we could either use the enduro.toml config file, or by adding dedicated config files - with a dedicated config file, we could use CSVs instead of TOML, which would be more user-friendly. Open to any option, however!

Keep it as is and cry every time a user forgets to reset their list.

**Additional context**

djjuhasz:
Fixes #60

- Add a "fileFormat" section to the preprocessing config file
- Add a "AllowlistPath" config option to the "fileFormat" section
- If no "AllowlistPath" is configured, then file format validation will be skipped
- Add an allowed format CSV file to the dev and enduro kube configs


djjuhasz:
Note: this PR depends on PR#62

hack/kube/overlays/dev/preprocessing-secret.yaml
```diff
@@ -47,3 +47,41 @@ stringData:
     secretKey = "minio123"
     region = "us-west-1"
     bucket = "ais"
+
+    [fileformat]
+    allowlistPath = "/home/preprocessing/.config/allowed_file_formats.csv"
```
jraddaoui, line 52: Should it be `allowListPath`?

internal/activities/validate_file_formats.go
```diff
@@ -2,75 +2,65 @@ package activities
...
-		"fmt/653":   {},
+	if a.cfg.AllowlistPath == "" {
+		logger.Info("ValidateFileFormats: No file format allowlist path set, skipping file format validation")
+
+		return nil, nil
+	}
+
+	f, err := os.Open(a.cfg.AllowlistPath)
+	if err != nil {
+		return nil, fmt.Errorf("ValidateFileFormats: %v", err)
```
jraddaoui, line 53: The `%v` vs. `%w` situation always makes me wonder. I think we were trying to use `%w` in most cases to be able to unwrap it if it's needed.

internal/activities/validate_file_formats.go
```diff
@@ -96,3 +86,34 @@ func (a *ValidateFileFormats) Execute(
...
+		row, err := cr.Read()
+		if err == io.EOF {
+			break
+		}
+		if err != nil {
+			return nil, fmt.Errorf("invalid CSV: %v", err)
+		}
+
+		// Skip the first row.
+		if i > 0 {
```
jraddaoui, line 105: The default configuration above doesn't have a header, should we remove this skip or expect a header in the CSV?

jraddaoui: On the other hand, the configuration from the testdata folder includes the header.

jraddaoui (changes requested):
Looks great @djjuhasz!

Now that it's a configurable activity, I wonder if this should be part of the Enduro workflow. I know the plan was to make it a shared activity after this PR, but we can discuss with the team if it should be a default in Enduro instead of part of custom child workflows.

hack/kube/overlays/dev/preprocessing-secret.yaml
djjuhasz: I'm using "allowlist" as a single word.  I think "allow list path" is confusing because it's not that "list" is being used as a noun and not a verb or an adjective describing "path".

internal/activities/validate_file_formats.go
djjuhasz: @sevein is a very strong proponent of `%v` in almost all cases.  My understanding is that `%w` shouldn't be used when you don't want to expose the underlying implementation of the error.  I don't see any reason we'd want to unwrap the `os.Open` error here.

internal/activities/validate_file_formats.go
djjuhasz: Oh, that's a bug, the  default config should have a header.

djjuhasz: Fixed.

internal/activities/validate_file_formats.go
```diff
@@ -96,3 +86,34 @@ func (a *ValidateFileFormats) Execute(
...
+			// Get the file format identifier from the first column of each row
+			// and ignore subsequent columns.
+			formats[row[0]] = struct{}{}
```
jraddaoui, line 108: Should we check for empty rows/columns?

jraddaoui (approved):

internal/activities/validate_file_formats.go
djjuhasz: @jraddaoui I added a test to confirm that an empty row in an allowlist is ignored and doesn't cause an error.

I also added an explicit check for a "PRONOM PUID" (any letter case) header to allow for other columns in the allowlist, and added an extra column to the sample and test allowlist CSVs.

jraddaoui: Thanks!

jraddaoui (approved):
LGTM!

PR #64, opened 2024-10-18
Use xmlvalidate to validate SIP manifests

### LINKED ISSUE #39: Feature: use the xmlvalidate activity and SIP XSD files to validate XML metadata

We are still using [a Python script](https://github.com/artefactual-sdps/preprocessing-sfa/blob/main/hack/sampledata/xsd/xsdval.py) and [a fixed local version of the XSD files](https://github.com/artefactual-sdps/preprocessing-sfa/tree/main/hack/sampledata/xsd) to validate the metadata.xml or UpdatedAreldaMetadata.xml files.

@mcantelon recently created an activity in the temporal-activities respository to perform that validation:

https://github.com/artefactual-sdps/temporal-activities/blob/main/xml/xsd_validate_activity.go

We also need to change the workflow to use the XSD files contained on each SIP instead of the fixed ones copied into the environment.

djjuhasz:
Fixes #39

Switch from the internal "validate metadata" activity which calls a
Python (xsdval.py) script to the `temporal-activities/xmlvalidate`
module. xmlvalidate calls the xmllint C program to validate the SIP
manifest file against the XSD schema files included in the SIP.

- Install xmllint in the preprocessing-sfa Docker image
- Import github.com/artefactual-sdps/temporal-activities/xmlvalidate
- Switch to xmlvalidate with the xmllint validator for validating the
  SIP metadata file
- Remove the internal "validate metadata" activivity
- Remove the sampledata directory containing the `xsdval.py` script,
  Arelda XSD files, and sample SIP


jraddaoui (changes requested):
Really nice, thanks @djjuhasz! Just one comment about the Docker image.

Dockerfile
```diff
@@ -22,22 +25,32 @@ RUN --mount=type=cache,target=/go/pkg/mod \
...
-USER root
-RUN apk add --update --no-cache python3 && \
-	ln -sf python3 /usr/bin/python && \
-	python3 -m ensurepip
 USER preprocessing
-RUN pip3 install --no-cache --upgrade pip lxml bagit==v1.8.1
-COPY --from=build-preprocessing-worker --link /src/hack/sampledata/xsd/* /
 COPY --from=build-preprocessing-worker --link /out/preprocessing-worker /home/preprocessing/bin/preprocessing-worker
+RUN mkdir /home/preprocessing/shared
+
```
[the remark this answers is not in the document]
djjuhasz: Huh, I think you're right that we don't use bagit-python in preprocessing-sfa anymore. I guess we only do bag validation in Enduro, and we're using [bagit-gython](https://github.com/artefactual-labs/bagit-gython) there.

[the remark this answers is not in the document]
jraddaoui: Yeah, we just forgot to remove it when changing to the shared activity that uses go-bagit.

Dockerfile
```diff
@@ -1,6 +1,6 @@
-# syntax = docker/dockerfile:1.4
+# syntax=docker/dockerfile:1
 
-ARG GO_VERSION
+ARG GO_VERSION=1.23.2
```
jraddaoui, line 3: I think it's better and easier to maintain without a default. Relying on [the bash script to build](https://github.com/artefactual-sdps/preprocessing-sfa/blob/main/hack/build_docker.sh) and the .go-version. Or, like the recent changes in Enduro, [reading from go.mod directly](https://github.com/artefactual-sdps/enduro/blob/main/hack/build_docker.sh#L45-L49).

djjuhasz: Not supplying a default value for a required ARG value goes against best practice: https://docs.docker.com/reference/build-checks/invalid-default-arg-in-from/ and the Docker linter I'm using complains about it.

djjuhasz: Adding a default value also doesn't change the results when building the Docker image from the bash script or in the Enduro case.

jraddaoui: I know, it's just sad to have to update the default, but I can see their point.

djjuhasz: Yeah, it is unfortunate that it's one more place we have to update the Go version. :(

jraddaoui (approved):
Really nice, thanks @djjuhasz!

PR #65, opened 2024-10-18
Reduce gosec verbosity in `make pre-commit`

djjuhasz:
Reduce the noise from gosec when running `make pre-commit`.


jraddaoui (approved):
:ok_hand:

PR #67, opened 2024-10-21
Don't rename Prozess_Digitalisierung_PREMIS.xml

### LINKED ISSUE #58: Problem: Prozess_Digitalisierung_PREMIS.xml should not append the parent directory name when moved to the metadata folder
**Describe the bug**

We thought it would be possible to include multiple dossiers in the same digitized SIP and appended the parent directory (dossier) name to avoid file name conflicts in the metadata directory.

**To Reproduce**

Ingest a digitized SIP type that includes a Prozess_Digitalisierung_PREMIS.xml file.

**Expected behavior**

Prozess_Digitalisierung_PREMIS.xml should be moved as it is to the metadata directory.

djjuhasz:
Fixes #58

Digitized (Vecteur) SIPs are the only type of SIP that contain a Prozess_Digitalisierung_PREMIS.xml file, and they will only ever contain a single dossier.  Because digitized SIPs only have a single dossier that contains a single Prozess_Digitalisierung_PREMIS.xml file there's no need to rename the file in the PIP to avoid name clashes.

Changes:
- Update the SIP identification logic - if a SIP has a Prozess_Digitalisierung_PREMIS.xml file then it must be a digitized SIP
- Add a SIP validation check to confirm that digitized SIPs only contain a single dossier
- Remove the SIP transformation code that renamed the Prozess_Digitalisierung_PREMIS.xml file

### DIFF
--- a/internal/activities/validate_structure.go
+++ b/internal/activities/validate_structure.go
@@ -74,6 +74,18 @@ func (a *ValidateStructure) Execute(
 		failures = append(failures, extras...)
 	}
 
+	// Check that digitized SIPs only have one dossier in the content dir.
+	if params.SIP.Type == enums.SIPTypeDigitizedSIP {
+		entries, err := os.ReadDir(params.SIP.ContentPath)
+		if err != nil {
+			return nil, fmt.Errorf("ValidateStructure: check for unexpected dossiers: %v", err)
+		}
+
+		if len(entries) > 1 {
+			failures = append(failures, "More than one dossier in the content directory")
+		}
+	}
+
 	return &ValidateStructureResult{Failures: failures}, nil
 }
 
--- a/internal/pips/pips.go
+++ b/internal/pips/pips.go
@@ -1,7 +1,6 @@
 package pips
 
 import (
-	"fmt"
 	"path/filepath"
 	"strings"
 
@@ -42,20 +41,16 @@ func (p PIP) Name() string {
 }
 
 func (p PIP) ConvertSIPPath(path string) string {
-	switch {
-	case filepath.Base(path) == "Prozess_Digitalisierung_PREMIS.xml":
-		parent := filepath.Base(filepath.Dir(path))
-		return filepath.Join(
-			"metadata",
-			fmt.Sprintf("Prozess_Digitalisierung_PREMIS_%s.xml", parent),
-		)
-	case filepath.Base(path) == "metadata.xml":
-		return filepath.Join("objects", p.Name(), "header", "metadata.xml")
-	case filepath.Base(path) == "UpdatedAreldaMetadata.xml":
-		return filepath.Join("metadata", "UpdatedAreldaMetadata.xml")
-	case strings.HasPrefix(path, "content"):
-		return filepath.Join("objects", p.Name(), path)
-	default:
-		return ""
+	switch name := filepath.Base(path); name {
+	case "Prozess_Digitalisierung_PREMIS.xml", "UpdatedAreldaMetadata.xml":
+		return filepath.Join("metadata", name)
+	case "metadata.xml":
+		return filepath.Join("objects", p.Name(), "header", name)
 	}
+
+	if strings.HasPrefix(path, "content") {
+		return filepath.Join("objects", p.Name(), path)
+	}
+
+	return ""
 }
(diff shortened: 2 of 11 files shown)


internal/activities/validate_structure.go
```diff
@@ -74,6 +74,22 @@ func (a *ValidateStructure) Execute(
...
+			return nil, fmt.Errorf("ValidateStructure: check for unexpected dossiers: %v", err)
+		}
+
+		if len(entries) > 1 {
+			for _, entry := range entries[1:] {
+				failures = append(failures, fmt.Sprintf(
+					"Unexpected dossier: %s", entry.Name(),
+				))
+			}
+		}
```
jraddaoui, line 90: We don't know here which ones are the unexpected, if you want to show the directory names I'd show them all. Maybe "More than one dossier folder in the content directory: [list of directories]", or something like that. I also think not showing the list of directories is okay.

internal/activities/validate_structure.go
```diff
@@ -74,6 +74,22 @@ func (a *ValidateStructure) Execute(
...
+		entries, err := os.ReadDir(params.SIP.ContentPath)
+		if err != nil {
+			return nil, fmt.Errorf("ValidateStructure: check for unexpected dossiers: %v", err)
+		}
+
+		if len(entries) > 1 {
+			for _, entry := range entries[1:] {
+				failures = append(failures, fmt.Sprintf(
+					"Unexpected dossier: %s", entry.Name(),
+				))
```
jraddaoui, line 88: This could go in one line. If not, I prefer to split each argument on its own line:

```go
failures = append(
    failures,
    fmt.Sprintf("Unexpected dossier: %s", entry.Name()),
)
```

internal/pips/pips.go
```diff
@@ -42,20 +41,18 @@ func (p PIP) Name() string {
...
-	case filepath.Base(path) == "UpdatedAreldaMetadata.xml":
-		return filepath.Join("metadata", "UpdatedAreldaMetadata.xml")
-	case strings.HasPrefix(path, "content"):
+	switch name := filepath.Base(path); name {
+	case "Prozess_Digitalisierung_PREMIS.xml":
+		return filepath.Join("metadata", name)
+	case "metadata.xml":
+		return filepath.Join("objects", p.Name(), "header", name)
+	case "UpdatedAreldaMetadata.xml":
+		return filepath.Join("metadata", name)
```
jraddaoui, line 50: `Prozess_Digitalisierung_PREMIS.xml` and `UpdatedAreldaMetadata.xml` could now go in the same case.

internal/sip/sip.go
```diff
@@ -55,7 +54,7 @@ func New(path string) (SIP, error) {
 	if err != nil {
 		return s, fmt.Errorf("SIP: New: %v", err)
 	}
-	if len(f) > 0 && strings.Contains(strings.ToLower(s.Path), "vecteur") {
+	if len(f) > 0 {
```
jraddaoui, line 57: Why are you removing this requirement?

jraddaoui (changes requested):
Nice @djjuhasz, thanks!

We also need to change it in the `premis` package: https://github.com/artefactual-sdps/preprocessing-sfa/blob/main/internal/premis/premis.go#L380-L384.

internal/sip/sip.go
djjuhasz: I removed it because it didn't seem important to me — If the SIP contains a "Prozess_Digitalisierung_PREMIS.xml" file then it must be a digitized SIP whether or not it has "vecteur" in the name. Conversely if a SIP has a "Prozess_Digitalisierung_PREMIS.xml" file but doesn't have "vecteur" in the same then it shouldn't be identified as "born digital" SIP in my opinion.

jraddaoui: @fiver-watson this was the requirement that checked for "vecteur" in the SIP name (that we made case insensitive). Is it okay to remove it? Not sure where it comes from.

djjuhasz: @fiver-watson I'm just waiting on your response to @jraddaoui before submitting this PR for a final code review.

jraddaoui (approved):

Issue #59, opened 2024-10-03
Feature: use relative paths instead of absolute paths in error messages


sallain:
**Is your feature request related to a problem? Please describe.**

When a user reviews an error message for a validation activity that failed, they are shown the absolute path of the problematic file or directory. The absolute path is long and much of it is meaningless to the average user; in talk-aloud testing, we found that they skipped right over it and were unable to identify which file or directory was causing the issue.

![image](https://github.com/user-attachments/assets/342011ac-f413-4e32-93bc-6bd1e4fa7307)

The absolute path is used in error messages in the following activities:

* _Validate SIP structure_
* _Validate SIP file formats_

On a related, but inverse, note, _Verify SIP manifest_ seems to be already using relative paths, though actually it might be TOO short. For the sake of consistency, all paths should start at the top-level SIP directory.

![image](https://github.com/user-attachments/assets/6e43709a-ab84-4212-914b-13d7b1031420)

**Describe the solution you'd like**

Using the relative path of the file, starting at the top-level directory of the SIP, should make the path more comprehensible to users.

**Describe alternatives you've considered**

Can't really think of any - we need to provide enough information that the file/directory is uniquely identifiable, but not too much. Hopefully this is the right balance.

**Additional context**

This is a child issue of #43

sallain:
LGTM!

fiver-watson:
@DanielCosme @sallain I see that feedback was added to this, but i am not sure it was noticed, so I am leaving this comment and moving this card back to in progress, so that the additional absolute paths found during QA above can be addressed at some point.

PR #71, opened 2024-10-25
Implement relative path to error SIP validation report

DanielCosme:
Solves: #59

### DIFF
--- a/internal/activities/validate_file_formats.go
+++ b/internal/activities/validate_file_formats.go
@@ -77,8 +77,14 @@ func (a *ValidateFileFormats) Execute(
 			return fmt.Errorf("identify format: %v", err)
 		}
 
+		rel, err := filepath.Rel(params.SIP.Path, p)
+		if err != nil {
+			return fmt.Errorf("get relative path: %v", err)
+		}
+		rel = filepath.Join(filepath.Base(params.SIP.Path), rel)
+
 		if _, ok := allowed[ff.ID]; !ok {
-			failures = append(failures, fmt.Sprintf("file format %q not allowed: %q", ff.ID, p))
+			failures = append(failures, fmt.Sprintf("file format %q not allowed: %q", ff.ID, rel))
 		}
 
 		return nil
--- a/internal/activities/validate_file_formats_test.go
+++ b/internal/activities/validate_file_formats_test.go
@@ -123,14 +123,12 @@ fmt/95,"PDF/A "
 			want: activities.ValidateFileFormatsResult{
 				Failures: []string{
 					fmt.Sprintf(
-						`file format %q not allowed: "%s/dir/file1.png"`,
+						`file format %q not allowed: "test_transfer/content/content/dir/file1.png"`,
 						"fmt/11",
-						invalidFormatsContentPath,
 					),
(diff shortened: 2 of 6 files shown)


djjuhasz:
@DanielCosme The "Validate metadata" activity is still returning an absolute path on this branch:
image

JSON output (Tilt):
```
{"Name":"Validate SIP metadata","Message":"Content error: metadata validation has failed: /home/preprocessing/shared/enduro2008167314/extract1349385218/digitized_SIP_extra_dossier/header/metadata.xml:139: element datei: Schemas validity error : Element '{[http://bar.admin.ch/arelda/v4}datei',](http://bar.admin.ch/arelda/v4%7Ddatei',) attribute 'id': '_cQ6sm5CChWVqtqmrWvne0W' is not a valid value of the atomic type '{[http://bar.admin.ch/arelda/v4}idDatei'.\n/home/preprocessing/shared/enduro2008167314/extract1349385218/digitized_SIP_extra_dossier/header/metadata.xml](http://bar.admin.ch/arelda/v4%7DidDatei'./n/home/preprocessing/shared/enduro2008167314/extract1349385218/digitized_SIP_extra_dossier/header/metadata.xml) fails to validate\n"
```

djjuhasz:
Verify SIP Manifest output looks good:
image

djjuhasz:
> @DanielCosme The "Validate metadata" activity is still returning an absolute path on this branch: image
> 
> JSON output (Tilt):
> 
> ```
> {"Name":"Validate SIP metadata","Message":"Content error: metadata validation has failed: /home/preprocessing/shared/enduro2008167314/extract1349385218/digitized_SIP_extra_dossier/header/metadata.xml:139: element datei: Schemas validity error : Element '{[http://bar.admin.ch/arelda/v4}datei',](http://bar.admin.ch/arelda/v4%7Ddatei',) attribute 'id': '_cQ6sm5CChWVqtqmrWvne0W' is not a valid value of the atomic type '{[http://bar.admin.ch/arelda/v4}idDatei'.\n/home/preprocessing/shared/enduro2008167314/extract1349385218/digitized_SIP_extra_dossier/header/metadata.xml](http://bar.admin.ch/arelda/v4%7DidDatei'./n/home/preprocessing/shared/enduro2008167314/extract1349385218/digitized_SIP_extra_dossier/header/metadata.xml) fails to validate\n"
> ```

@DanielCosme oops, nevermind --- I got this branch confused with the "relatives-path" branch in the enduro repo.  Looks okay with this branch:
image

jraddaoui (approved):
Thanks @DanielCosme, LGTM!

It would be nice to move this to a SIP method `SIP.GetRelPath(path)`, but seeing how the paths come from the manifest and how that activity works, it may be even worst. I think this is okay for a first pass, if we find that we need to get the relatives path more often we can create that method and change how that activity works.

DanielCosme:
@jraddaoui The challenge with these error messages is that every activity does it's own thing. So when creating errors, the formatting and the actual thing that checks for them are to tightly coupled. So I could not find one Nice global way of doing this. Not sure why these activities where implemented like this, but that was the code I was given to work with so I did the least amount of changes I could possible do, to avoid regressions.

PR #75, opened 2024-11-04
Improve processing events interface

djjuhasz:
Add a few new processing event helper methods and update the existing methods to reduce repetitive code and further standardize how events are created.

### DIFF
--- a/internal/workflow/event_wrapper.go
+++ b/internal/workflow/event_wrapper.go
@@ -11,7 +11,7 @@ type eventWrapper struct {
 	*eventlog.Event
 }
 
-func newEvent(ctx temporalsdk_workflow.Context, name string) *eventWrapper {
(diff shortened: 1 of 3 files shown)


internal/eventlog/eventlog.go
```diff
@@ -34,3 +34,7 @@ func (e *Event) Complete(t time.Time, outcome enums.EventOutcome, msg string, a
 
 	return e
 }
+
+func (e *Event) Succeed(t time.Time, msg string, a ...any) *Event {
+	return e.Complete(t, enums.EventOutcomeSuccess, msg, a...)
+}
```
jraddaoui, line 40: I think this is not used, right? With the wrapper you always call `w.Event.Complete`.

jraddaoui (approved):
Looks good, thanks @djjuhasz!

PR #76, opened 2024-11-05
Fail CI when golines rewraps long lines

djjuhasz:


jraddaoui (approved):
Thanks @djjuhasz!

I took a different approach in https://github.com/artefactual-sdps/temporal-activities/commit/c5b2918daa1371186f4ac5753a1c9012d240feb9, where I also found it was needed in the `golangci-lint` job. Your solution looks better to me, I didn't know about `--dry-run`, but we should normalize in all projects and double check the solution for  the `golangci-lint` job.

jraddaoui:
Looking at [the failed job output](https://github.com/artefactual-sdps/preprocessing-sfa/actions/runs/11689583417/job/32552640741?pr=76), I think the solution in `temporal-activities` shows a better output, where you can see the actual changes needed. Just a thought, I still find this cleaner.

djjuhasz:
Thanks @jraddaoui.  I should have put a WIP tag on this PR. :laughing: I was just trying to test that it actually works and I'm not sure how else to trigger a Github workflow besides creating a PR.  I'll check out your suggestions.

PR #78, opened 2024-11-07
Update AIS integration

jraddaoui:
Update AIS integration to use Enduro's poststorage child workflows.

- Remove AIS API server.
- Remove `AIPName` and make `AIPUUID` a string in AIS `WorkflowParams`.
- Add local activity to get the AIP current path from AMSS.
- Parse AIP directory name from current path.

Related to https://github.com/artefactual-sdps/enduro/issues/886 and https://github.com/artefactual-sdps/enduro/pull/1061.


internal/ais/policies.go
```diff
@@ -0,0 +1,54 @@
...
+				"TemporalTimeout:StartToClose",
+			},
+		},
+	})
+}
+
+// withActivityOptsForLocalAction returns a workflow context with activity
+// options suited for local activities like disk operations that should not
+// require a retry policy attached.
+func withActivityOptsForLocalAction(ctx temporalsdk_workflow.Context) temporalsdk_workflow.Context {
```
djjuhasz, line 33: I've always found this name to be too similar to `withLocalActivityOpts`.  Suggested alternatives:

- withShortLivedActivityOpts
- withFilesystemActivityOpts
- withActivityOpts (assuming it's the default option)

internal/ais/workflow.go
```diff
@@ -47,6 +44,23 @@ func (w *Workflow) Execute(ctx temporalsdk_workflow.Context, params *WorkflowPar
 	logger := temporalsdk_workflow.GetLogger(ctx)
 	logger.Debug("AIS workflow running!", "params", params)
 
+	defer func() {
+		logger.Debug("AIS workflow finished!", "result", r, "error", e)
```
djjuhasz, line 48: This could be misleading — if the workflow returns an error the returned result will *always* be nil. I've been tricked by this weird Temporal quirk a few times.

internal/ais/workflow.go
```diff
@@ -154,10 +168,10 @@ func (w *Workflow) SessionHandler(ctx temporalsdk_workflow.Context, params *Work
 	if parseResult.UpdatedAreldaMetadataRelPath != "" {
 		var fetchMetadataResult FetchActivityResult
 		e = temporalsdk_workflow.ExecuteActivity(
```
djjuhasz, line 170: It seems inefficient to download the metadata files from the AMSS when we have a local copy in the SIP. We aren't making any changes to the `UpdatedAreldaMetadata.xml` or `metadata.xml` files in preservation are we?  Maybe we can pass the path of the relevant metadata file from Enduro to this workflow?

internal/ais/workflow.go
```diff
@@ -154,10 +168,10 @@ func (w *Workflow) SessionHandler(ctx temporalsdk_workflow.Context, params *Work
 	if parseResult.UpdatedAreldaMetadataRelPath != "" {
```
djjuhasz, line 168: Instead of this conditional why not do something like:
```go
var metadataRelPath string
if parseResult.UpdatedAreldaMetadataRelPath != "" {
    metadataRelPath = parseResult.UpdatedAreldaMetadataRelPath 
} else {
    metadataRelPath = parseResult.MetadataRelPath
}

if metadataRelPath != "" {
    var fetchMetadataResult FetchActivityResult
    e = temporalsdk_workflow.ExecuteActivity(
    ...
}
```

That last conditional is probably unnecessary, we should never have a SIP without one or the other metadata file.

djjuhasz (approved):
Thanks @jraddaoui.  I made a few comments, but I don't see any deal breakers.

internal/ais/workflow.go
jraddaoui: A different workflow and maybe a different worker, running in a different place than preprocessing/Enduro, so we don't have access to the SIP.

But you are right. As next steps I was planning to investigate the option of generating a preprocessing identifier that could be returned in the preprocessing child workflow result and added to the poststorage workflow params. That way we could temporary upload the `UpdatedAreldaMetadata.xml` or `metadata.xml` in MinIO from preprocessing and fetch it from this workflow using that identifier. It will prevent a second file extraction from the AIP and it also may remove the need of parsing the METS file. But that will be in different PRs, probably after the work you are doing to merge the metadata files together.

internal/ais/workflow.go
jraddaoui: Good call! I'll change the last condition and return an error if no metadata file was found.

internal/ais/workflow.go
jraddaoui: Right, I'll make sure we also return a nil result if there is an error.

internal/ais/policies.go
jraddaoui: Renamed it to `withFilesystemActivityOpts`.

PR #80, opened 2024-11-13
Add an activity to merge AIS metadata files

### LINKED ISSUE #77: Feature: combine METS and metadata files for delivery to AIS

DPS must deliver both the METS file and the metadata.xml/UpdatedAreldaMetadata.xml file to the AIS during the post-preservation workflow. However, AIS only expects one file.

Combine the METS and the metadata.xml/UpdatedAreldaMetadata.xml files together into one metadata file. For migration files (files identified as DigitizedAIP or BornDigitalAIP), UpdatedAreldaMetadata.xml should be used.

The newly created file should be named with the prefix `AIS_` followed by the accession number, which can be found in the metadata.xml (or UpdatedAreldaMetadata.xml, but should be the same value) under `<ablieferungsnummer>`. There should only be one ablieferungsnummer per metadata file. The number is formatted as `2002/05`; the / should be replaced with an _. The final file name will be `AIS_2002_05`.

Within the file, SFA would like the contents of metadata.xml/UpdatedAreldaMetadata.xml first, since it contains the higher hierarchies, and then the METS. The contents of the two files should probably be tagged in some way but I think it can be pretty simple - perhaps just indicating the source file.

None

**Additional context**

There's a very real chance that, when operating at scale, the resulting file will be too big for AIS to handle; it might make sense to then limit which fields from each file we're combining into this new file. But we'll tackle that if/when it happens.

djjuhasz:
Fixes #77.

Concatenate the Arelda metadata file from the original package and the METS file created by Archivematica into a single "AIS" metadata file.

### DIFF
--- a/internal/ais/combinemd.go
+++ b/internal/ais/combinemd.go
@@ -1,0 +1,135 @@
+package ais
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"io"
+	"os"
+	"path/filepath"
+	"strings"
+
+	"github.com/antchfx/xmlquery"
+
+	"github.com/artefactual-sdps/preprocessing-sfa/internal/fsutil"
+)
+
+const CombineMDActivityName = "combine-metadata-files"
+
+type (
+	CombineMDActivity       struct{}
+	CombineMDActivityParams struct {
+		AreldaPath string
+		METSPath   string
+		LocalDir   string
+	}
+	CombineMDActivityResult struct {
+		Path string
+	}
+)
+
+func NewCombineMDActivity() *CombineMDActivity {
+	return &CombineMDActivity{}
+}
+
+func (a *CombineMDActivity) Execute(
+	ctx context.Context,
+	params CombineMDActivityParams,
+) (*CombineMDActivityResult, error) {
+	if !fsutil.FileExists(params.AreldaPath) {
+		return nil, fmt.Errorf("missing Arelda file: %s", params.AreldaPath)
+	}
+	if !fsutil.FileExists(params.METSPath) {
+		return nil, fmt.Errorf("missing METS file: %s", params.METSPath)
+	}
+
+	name, err := aisFilename(params.AreldaPath)
+	if err != nil {
+		return nil, fmt.Errorf("name AIS file: %v", err)
+	}
+
+	dest := filepath.Join(params.LocalDir, name)
+
+	// Combine metadata files into AIS file.
+	w, err := os.Create(dest) // #nosec G304 -- generated path.
+	if err != nil {
+		return nil, fmt.Errorf("create AIS file: %v", err)
+	}
+	defer w.Close()
+
+	if err := w.Chmod(os.FileMode(0o644)); err != nil {
+		return nil, fmt.Errorf("set AIS file permissions: %v", err)
+	}
+
+	if err = concat(w, filepath.Join(params.AreldaPath), filepath.Join(params.METSPath)); err != nil {
+		return nil, fmt.Errorf("concat: %v", err)
+	}
+
+	// Delete original metadata files.
+	if err = removePaths(params.AreldaPath, params.METSPath); err != nil {
+		return nil, fmt.Errorf("removePaths: %v", err)
+	}
+
+	return &CombineMDActivityResult{Path: dest}, nil
+}
+
+func aisFilename(mdpath string) (string, error) {
+	id, err := parseAccessionID(mdpath)
+	if err != nil {
+		return "", fmt.Errorf("get accession number: %v", err)
+	}
+
+	id = strings.ReplaceAll(id, "/", "_")
+
+	return fmt.Sprintf("AIS_%s", id), nil
+}
+
+func parseAccessionID(path string) (string, error) {
+	f, err := os.Open(path) // #nosec G304 -- trusted path.
+	if err != nil {
+		return "", fmt.Errorf("open metadata file: %v", err)
+	}
+	defer f.Close()
+
+	sp, err := xmlquery.CreateStreamParser(f, "//paket/ablieferung/ablieferungsnummer")
+	if err != nil {
+		return "", fmt.Errorf("create XML parser: %v", err)
+	}
+
+	n, err := sp.Read()
+	if err == io.EOF {
+		return "", fmt.Errorf("can't find ablieferungsnummer in %q", filepath.Base(path))
+	}
+	if err != nil {
+		return "", fmt.Errorf("read XML stream: %v", err)
+	}
+	return n.InnerText(), nil
+}
+
+func concat(w io.Writer, paths ...string) error {
+	for i := range paths {
+		r, err := os.Open(paths[i]) // #nosec G304 -- trusted path.
+		if err != nil {
+			return fmt.Errorf("read: %v", err)
+		}
+		defer r.Close()
+
+		if _, err := io.Copy(w, r); err != nil {
+			return fmt.Errorf("copy: %v", err)
+		}
+		_ = r.Close()
+	}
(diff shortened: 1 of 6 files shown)


djjuhasz:
@jraddaoui I haven't been able to test this locally because I don't know how to configure the AIS worker, but I'm open the PR anyway to get an initial code review.

internal/ais/combinemd.go
```diff
@@ -0,0 +1,138 @@
...
+)
+
+const CombineMDActivityName = "combine-metadata-files"
+
+type (
+	CombineMDActivity       struct{}
+	CombineMDActivityParams struct {
+		AreldaRelPath string
+		METSRelPath   string
+		WorkingDir    string
```
jraddaoui, line 24: Maybe call it `localDir` like in the workflow, `workingDir` (from the config) would be the parent directory.

internal/ais/combinemd.go
```diff
@@ -0,0 +1,138 @@
...
+func (a *CombineMDActivity) Execute(
+	ctx context.Context,
+	params CombineMDActivityParams,
+) (*CombineMDActivityResult, error) {
+	areldaPath := filepath.Join(params.WorkingDir, params.AreldaRelPath)
+	if !fsutil.FileExists(areldaPath) {
+		return nil, fmt.Errorf("missing Arelda file: %s", areldaPath)
+	}
+
+	metsPath := filepath.Join(params.WorkingDir, params.METSRelPath)
```
jraddaoui, line 44: Same.

internal/ais/workflow.go
```diff
@@ -188,6 +188,20 @@ func (w *Workflow) SessionHandler(ctx temporalsdk_workflow.Context, aipUUID, aip
 		return "", e
 	}
 
+	var combineMDResult CombineMDActivityResult
+	e = temporalsdk_workflow.ExecuteActivity(
+		withFilesystemActivityOpts(ctx),
+		CombineMDActivityName,
+		&CombineMDActivityParams{
+			AreldaRelPath: metadataRelPath,
```
jraddaoui, line 196: The absolute path is the destination used in the previous activity params.

internal/ais/workflow.go
```diff
@@ -188,6 +188,20 @@ func (w *Workflow) SessionHandler(ctx temporalsdk_workflow.Context, aipUUID, aip
...
+			METSRelPath:   metsPath,
```
jraddaoui, line 197: This is already an absolute path.

internal/ais/workflow.go
```diff
@@ -188,6 +188,20 @@ func (w *Workflow) SessionHandler(ctx temporalsdk_workflow.Context, aipUUID, aip
...
+			WorkingDir:    filepath.Join(localDir, filepath.Base(metadataRelPath)),
```
jraddaoui, line 198: Just use `localDir`, I think we don't need an extra folder with the name of the metadata file.

internal/testdata/little-Test-AIP-Digitization/AIS_1000_893_3251903
```diff
@@ -0,0 +1,748 @@
...
+   <archivischerVorgang>
+      <vorgangstyp>Eingangsprüfung manuell</vorgangstyp>
+      <beschreibung>SIP Approved for Ingest</beschreibung>
+      <datum>
+         <von>2023-08-31T15:36:23.236+02:00</von>
+         <bis>2023-08-31T15:36:23.236+02:00</bis>
+      </datum>
+      <bearbeiter>Andreas Christe</bearbeiter>
+   </archivischerVorgang>
+</paket>
```
jraddaoui, line 748: Is this needed in the end?

jraddaoui (changes requested):
Really nice, thanks for that workflow test @djjuhasz! 😍

I think it would be better to pass the absolute paths directly instead of splitting and then re-joining them in the activity. I'd create a variable in the workflow for the metadata file absolute path, and use that as the `Destination` in the fetch activity and as the `AreldaPath` in the combine activity. Other than that and removing the test data file, it looks great to me.

internal/ais/combinemd.go
```diff
@@ -0,0 +1,138 @@
...
+	"github.com/artefactual-sdps/preprocessing-sfa/internal/fsutil"
+)
+
+const CombineMDActivityName = "combine-metadata-files"
+
+type (
+	CombineMDActivity       struct{}
+	CombineMDActivityParams struct {
+		AreldaRelPath string
+		METSRelPath   string
```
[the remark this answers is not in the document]
djjuhasz: I'm not clear if you are saying "we could pass the absolute paths, if we want to" or "we should be using absolute paths here".

internal/ais/combinemd.go
jraddaoui: Feel free to ignore.

internal/ais/combinemd.go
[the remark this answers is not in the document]
djjuhasz: Using absolute paths makes the tests a bit harder, if nothing else. It also duplicates the path information in the "WorkingDir" parameter.

internal/ais/workflow.go
djjuhasz: Again, I found the paths and naming convention used in the workflow very unclear.

internal/ais/combinemd.go
jraddaoui: Working dir is the config value, used by the worker for all workflows. Local dir is the specific one created for this workflow. Definitely not clear.

internal/ais/combinemd.go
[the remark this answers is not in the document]
jraddaoui: I think it's better to use absolute paths since we already have them in the workflow (`Destination` params used in the fetch activities), it seems unnecessary to me to recreate them in the activity.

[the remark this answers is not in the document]
djjuhasz: Okay, I'll switch to absolute paths.

internal/ais/workflow.go
jraddaoui: I agree, it would be nice to add a state struct where we can add some comments about the variables. An improvement for another PR.

internal/ais/combinemd.go
djjuhasz: Hmm, looking at this again the name of `localDir` is deterministic and will be the same for multiple workflow runs if the workingDir is the same (which is always true for a single worker), and the AIP UUID is the same.  If two workflows got started with the same UUID it could cause file lock or race conditions.

It's probably not a problem right now because we are limiting the number of concurrent running workflows to one, but if we ever do increase the number of workflows the worker can simultaneously process it could be an issue.

internal/testdata/little-Test-AIP-Digitization/AIS_1000_893_3251903
djjuhasz: Good catch. Deleted.

internal/ais/combinemd.go
jraddaoui: That's a great point and something worth looking at in detail, even if you deploy two workers using the same working directory.

When we [start these workflows](https://github.com/artefactual-sdps/enduro/blob/main/internal/workflow/processing.go#L1136-L1141) from Enduro we also use the AIP UUID for the workflow ID, but we don't set a [`WorkflowIDReusePolicy`](https://pkg.go.dev/go.temporal.io/api/enums/v1#WorkflowIdReusePolicy) and I'm not sure what would happen if unspecified. This could be configurable in Enduro, but we still need to consider all options in the workflow.

jraddaoui (approved):
Thanks @djjuhasz!

PR #81, opened 2024-11-15
Minor fixes to AIS workflow

djjuhasz:
Refs #77.

- Remove extraneous `filepath.Join()` calls
- Improve commentary a bit
- Correct "search_md" zip name in workflow tests


jraddaoui (approved):
:ok_hand:

PR #83, opened 2024-11-15
Switch to ffvalidate for file format validation

djjuhasz:
Switch to https://github.com/artefactual-sdps/temporal-activities/tree/main/ffvalidate for file format identification and validation.


jraddaoui (approved):
Nice, thanks @djjuhasz!

PR #87, opened 2024-11-20
Remove SIP path from SIP metadata validation

DanielCosme:
This MR will remove the absolute path from the `Validate SIP metadata activity`. 

Continuation of: https://github.com/artefactual-sdps/preprocessing-sfa/issues/59#issuecomment-2460199726

Previously:   
image

Now:   
image

### DIFF
--- a/internal/workflow/preprocessing.go
+++ b/internal/workflow/preprocessing.go
@@ -219,11 +219,10 @@ func (w *PreprocessingWorkflow) Execute(
 	}
 
 	if validateMetadata.Failures != nil {
-		escapedFailures := make([]string, len(validateMetadata.Failures))
-		for _, f := range validateMetadata.Failures {
-			escapedFailures = append(escapedFailures, strings.ReplaceAll(f, identifySIP.SIP.Path, ""))
+		for idx, f := range validateMetadata.Failures {
+			validateMetadata.Failures[idx] = strings.ReplaceAll(f, identifySIP.SIP.Path+"/", "")
 		}
-		result.validationError(ctx, ev, "metadata validation has failed", escapedFailures)
+		result.validationError(ctx, ev, "metadata validation has failed", validateMetadata.Failures)
 	} else {
 		ev.Succeed(ctx, "Metadata validation successful")
 	}
--- a/internal/workflow/preprocessing.go
+++ b/internal/workflow/preprocessing.go
@@ -219,7 +219,11 @@ func (w *PreprocessingWorkflow) Execute(
 	}
 
 	if validateMetadata.Failures != nil {
-		result.validationError(ctx, ev, "metadata validation has failed", validateMetadata.Failures)
+		escapedFailures := make([]string, len(validateMetadata.Failures))
+		for _, f := range validateMetadata.Failures {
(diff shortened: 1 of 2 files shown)


internal/workflow/preprocessing.go
```diff
@@ -219,7 +219,11 @@ func (w *PreprocessingWorkflow) Execute(
 	}
 
 	if validateMetadata.Failures != nil {
-		result.validationError(ctx, ev, "metadata validation has failed", validateMetadata.Failures)
+		escapedFailures := make([]string, len(validateMetadata.Failures))
+		for _, f := range validateMetadata.Failures {
+			escapedFailures = append(escapedFailures, strings.ReplaceAll(f, identifySIP.SIP.Path, ""))
```
djjuhasz, line 224: Because you've already allocated the elements of the `escapedFailures` slice, you don't want to use append here as it will make the slice length = len(validateMetadata.Failures) * 2. E.g. https://go.dev/play/p/hWlL9IMZ9sA.

Instead you can do:
```
for i, f := range validateMetadata.Failures {
	escapedFailures[i] = strings.ReplaceAll(f, identifySIP.SIP.Path, "")
}
```

internal/workflow/preprocessing_test.go
```diff
@@ -613,7 +613,7 @@ file format fmt/11 not allowed: "fake/path/to/sip/file2.png"`,
 				},
 				{
 					Name:        "Validate SIP metadata",
-					Message:     "Content error: metadata validation has failed:\nfake/path/to/sip/additional/UpdatedAreldaMetadata.xml does not match expected metadata requirements",
+					Message:     "Content error: metadata validation has failed:\n\n/additional/UpdatedAreldaMetadata.xml does not match expected metadata requirements",
```
djjuhasz, line 616: I think the extra linebreak is due to my comment on using `append()`.

djjuhasz (changes requested):
Thanks @DanielCosme, just one note to address.

internal/workflow/preprocessing.go
jraddaoui, line 224: We should remove the initial slash too, so it looks like a relative path. Also for consistency with the changes in https://github.com/artefactual-sdps/preprocessing-sfa/pull/71.

DanielCosme: Makes sense. Now that I think about it, how do you feel about this? @djjuhasz 
```go
		for idx, f := range validateMetadata.Failures {
			validateMetadata.Failures[idx] = strings.ReplaceAll(f, identifySIP.SIP.Path, "")
		}
		result.validationError(ctx, ev, "metadata validation has failed", validateMetadata.Failures)
```

DanielCosme: Now new slice.

internal/workflow/preprocessing_test.go
DanielCosme: Yeah.

internal/workflow/preprocessing.go
DanielCosme: Done 8ec4c63f8c8449e425f955b9602aa912b47a2fe5

djjuhasz (approved):
👍🏻

Issue #88, opened 2024-11-20
Problem: "Verify SIP checksums" activity assumes checksums will be MD5


sallain:
**Describe the bug**

The checksum validation activity assumes that checksums in the provided metadata will be MD5. However, the eCH-0160 standard states that MD5, SHA-256, SHA-512, and SHA-1 are all permissible. Some of the sample packages that we have use SHA-256, and therefore fail on checksum validation.

**To Reproduce**

1. Upload a sample package with non-MD5 checksums (e.g. GEVER)
2. See it fail!

**Expected behavior**

The activity should support all valid checksum algorithms. In the metadata file, the checksum's algorithm is defined:

```
<pruefalgorithmus>MD5</pruefalgorithmus>
<pruefsumme>aa1f7ad13064e1643ac85478296165af</pruefsumme>
```

```
<pruefalgorithmus>SHA-256</pruefalgorithmus>
<pruefsumme>F4F4789FD15E5B9B19E0F2C2D640AC2BEA0690424FFD9258E6B35028A1981D0B</pruefsumme>
```

The activity should check the `pruefalgorithmus` tag to determine the type of checksum and check using the correct algorithm.

**Screenshots**

**Additional context**

sallain:
Ran a package with SHA-256 checksums with no problems!

sallain:
I also ran a package with mixed checksums (MD5, SHA-1, SHA-256, and SHA-512) and it also worked

PR #89, opened 2024-11-25
Support more checksum algorithms (#88)

mcantelon:
Added support for SHA-1, SHA-256, and SHA-512 file checksums.

Refs #88.

### DIFF
--- a/internal/activities/verify_manifest.go
+++ b/internal/activities/verify_manifest.go
@@ -2,9 +2,13 @@ package activities
 
 import (
 	"context"
-	"crypto/md5" // #nosec: 501 -- not used for security.
+	"crypto/md5"    // #nosec: 501 -- not used for security.
+	"crypto/sha1"   // #nosec: 501 -- not used for security.
+	"crypto/sha256" // #nosec: 501 -- not used for security.
+	"crypto/sha512" // #nosec: 501 -- not used for security.
 	"encoding/hex"
 	"fmt"
+	"hash"
 	"io"
 	"io/fs"
 	"os"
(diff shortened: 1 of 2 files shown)


internal/activities/verify_manifest.go
```diff
@@ -199,6 +202,54 @@ func verifyChecksums(
...
+				failures = append(
+					failures,
+					fmt.Sprintf(
+						"Checksum mismatch for %q (expected: %q, got: %q)",
+						path,
+						file.Checksum.Hash,
+						hash,
+					),
+				)
+			}
```
jraddaoui, line 252: Early feedback, in case you didn't think about it already. Putting the switch in a single `generateHash` function reduces duplication:

```go
// Generate checksum from filesystem file contents.
hash, err := generateHash(filepath.Join(root, path), file.Checksum.Algorithm)
if err != nil {
	return nil, fmt.Errorf("generate %s hash: %v", file.Checksum.Algorithm, err)
}
if hash != file.Checksum.Hash {
	failures = append(
		failures,
		fmt.Sprintf(
			"Checksum mismatch for %q (expected: %q, got: %q)",
			path,
			file.Checksum.Hash,
			hash,
		),
	)
}
```

```go
func generateHash(path, alg string) (string, error) {
	var h hash.Hash
	switch alg {
	case "MD5":
		h = md5.New() // #nosec: G401 -- not used for security.
	case "SHA-1":
		h = sha1.New() // #nosec: G401 -- not used for security.
	case "SHA-256":
		h = sha256.New() // #nosec: G401 -- not used for security.
	case "SHA-512":
		h = sha512.New() // #nosec: G401 -- not used for security.
	default:
		return "", fmt.Errorf("hash algorithm %q is not supported", alg)
	}

	f, err := os.Open(path) // #nosec: G304 -- trusted path.
	if err != nil {
		return "", fmt.Errorf("open file: %v", err)
	}
	defer f.Close()

	if _, err := io.Copy(h, f); err != nil {
		return "", fmt.Errorf("copy contents: %v", err)
	}

	return hex.EncodeToString(h.Sum(nil)), nil
}
```

jraddaoui (approved):
Nice, thanks @mcantelon!

Issue #63, opened 2024-10-17
Feature: validate file formats against format specification


sallain:
**Is your feature request related to a problem? Please describe.**

SFA accepts a limited number of formats (enforced through the use of a format allow list), and requires those formats to be well-formed and valid according to the format specification. Currently, staff are responsible for checking format validation manually, using the [KOST-Val tool suite](https://kost-ceco.ch/cms/kost-val.html). SFA staff currently check the following formats using KOST-VAL:

* TIFF (fmt/353)
* PDF/A (fmt/95, fmt/354, fmt/476, fmt/477, fmt/478)
* JP2 (x-fmt/392)

They also validate SIARD (fmt/161, fmt/1196, fmt/1777) using a custom script. Other formats on their allow list that could be validated include WAV, Matroska, and MPEG-4.

The manual check is time-consuming for staff, only carried out in some cases, and the result is not captured for preservation.

**Describe the solution you'd like**

Implement an Ingest activity to automate this check. The activity should always validate designated formats against their specification and should also write a PREMIS event to the premis.xml file.

Approximately 90% of SFA's corpus is PDF/A. In the interest of getting the best possible validation results for the most files, I'd suggest that we implement [veraPDF](https://verapdf.org/) first.

**Describe alternatives you've considered**

Use KOST-Val (also a Java application; also it contains some licensed tools, like pdfaPilot)

Based on preliminary research ([see board notes here](https://miro.com/app/board/uXjVMlYiVgs=/?moveToWidget=3458764602603816307&cot=14)) it seems like [JHOVE](https://jhove.openpreservation.org/) can validate TIFF and JP2, as well as WAV, which may also make it a good option for an initial proof of concept, or for a second iteration of this feature.

**Additional context**

djjuhasz:
VeraPDF provides two "apps" - a [Java GUI](https://docs.verapdf.org/gui/) and [CLI](https://docs.verapdf.org/cli/). From the screenshot it looks like the Java GUI requires a windowing system (e.g. X11) and is built for human interaction. For our purposes the CLI seems like the best fit.

I'd rather not install veraPDF in our Enduro worker containers and then call it from the enduro workers.  While a local installation may be the simplest approach to implement it bloats our worker container with the Java RE and the veraPDF code, and doesn't follow the Docker best practice of [decoupling applcations](https://docs.docker.com/build/building/best-practices/#decouple-applications).

[...]

PR #92, opened 2024-11-27
Add VeraPDF validation of PDF/A files

djjuhasz:
Add the VeraPDF Java CLI application to the Enduro worker container and use it to validate SIP files identified as a PDF/A format.

Refs #63.


djjuhasz:
I've tested the VeraPDF validation in my local dev environment and it works, but all of the PDF/A files in  our sample SIP fail validation:

```json
[
  {
    "Failures": [
      "invalid PDF/A: \"content/content/d0003/d0004/p0006.pdf\"",
      "invalid PDF/A: \"content/content/d0003/d0004/p0007.pdf\"",
      "invalid PDF/A: \"content/content/d0001/p0001.pdf\"",
      "invalid PDF/A: \"content/content/d0001/p0002.pdf\"",
      "invalid PDF/A: \"content/content/d0001/p0003.pdf\"",
      "invalid PDF/A: \"content/content/d0002/p0004.pdf\"",
      "invalid PDF/A: \"content/content/d0002/p0005.pdf\""
    ]
  }
]
```

VeraPDF is also quite slow - it took almost 5 seconds to validate the 7 PDF/A files listed above. :open_mouth:

djjuhasz:
I've done some testing and almost all of the time used by VeraPDF was startup overhead (probably JRE startup time). When I pass VeraPDF a directory path instead of individual file paths it's much faster (just under 1 second for the 7 test PDFs) and it reports about 17 milliseconds to validate each PDF.

internal/fvalidate/verapdf_validator.go
```diff
@@ -0,0 +1,76 @@
...
+	return "veraPDF"
+}
+
+func (v *veraPDFValidator) Validate(path string) (string, error) {
+	// If the veraPDF cmd path is not set then skip validation.
+	if v.cmd == "" {
+		return "", nil
+	}
+
+	cmd := exec.Command(v.cmd, "--recurse", path) // #nosec: G204 -- trusted path.
```
djjuhasz, line 51: We may want to add allow configuring additional args (like "--flavour" or "--defaultFlavour") in `fvalidate.Config`, but it might also be an additional security risk.  I'm not sure if there's a reasonable way to sanitize `cmd` or if it's reasonable to just trust that the config isn't malicious.

internal/fvalidate/verapdf_validator.go
```diff
@@ -0,0 +1,76 @@
...
+	_, err := cmd.Output()
+	if err != nil {
+		switch e := err.(type) {
```
jraddaoui, line 54: We are only considering one case in this switch, I think it would be easier to read as part of the same if. The default case is the same as the else.

jraddaoui (approved):
Really nice, thanks @djjuhasz! LGTM, just a few minor details.

internal/activities/validate_files.go
```diff
@@ -0,0 +1,129 @@
+// ValidateFiles validates SIP files against a file format specification. The
+// only format validator currently implemented is PDF/A.
```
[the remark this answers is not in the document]
djjuhasz: I meant to document the file, not the package. I can move the comment to the method if you think that's clearer.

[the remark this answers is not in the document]
jraddaoui: Yeah, I think that's better. I'm not sure if that comment will appear describing the package somewhere.

internal/fvalidate/verapdf_validator.go
djjuhasz: All four of the returns in this logic tree are different, maybe this makes the logic branches clearer?

```
	if err != nil {
		if e, ok := err.(*exec.ExitError); ok {
			switch code := e.ExitCode(); code {
			case 1:
				// Exit code 1 indicates a validation error, and there is no
				// STDERR. In this case the full validation report is written to
				// STDOUT, but we are ignoring it right now because it is very
				// long.
				return "One or more PDF/A files are invalid", nil

			default:
				// Other exit codes (e.g. file not found) should write an error
				// message to STDERR.
				v.logger.Info("veraPDF validate", "exit code", e.ExitCode(), "STDERR", string(e.Stderr))
				return "", errors.New("PDF/A validation failed with an application error")
			}
		}
		return "", err
	}

	return "", nil
```

internal/fvalidate/verapdf_validator.go
djjuhasz: To be clear I was wondering about adding a way to specify general arguments for verapdf in the config file, something like:

```
args := append(cfg.args, path) 
cmd := exec.Command(v.cmd, "--recurse", args...)
```

Where `cfg.args` would be a slice of string arguments.

I'm not sure how adding a way to add arguments in config would add a version checking requirement?

internal/fvalidate/verapdf_validator.go
djjuhasz: I couldn't find any formal rule or guide in our developer resources but I think our practice is to usually return an explicit `nil` error from a function on the "happy" branch where no errors are encountered.  I think in a case like this returning `return "", err` where `err` can be nil or have an error value makes it harder to distinguish when an error is returned and when it isn't.

In this particular case, it might actually be better to return early if `err` is nil to avoid unnecessary nesting:

```
if err == nil { // error IS nil.
   return "", nil
}

if e, ok := err.(*exec.ExitError); !ok {
	return "", err
}

switch code := e.ExitCode(); code {
case 1:
	// Exit code 1 indicates a validation error, and there is no
	// STDERR. In this case the full validation report is written to
	// STDOUT, but we are ignoring it right now because it is very
	// long.
	return "One or more PDF/A files are invalid", nil

default:
	// Other exit codes (e.g. file not found) should write an error
	// message to STDERR.
	v.logger.Info("veraPDF validate", "exit code", e.ExitCode(), "STDERR", string(e.Stderr))
	return "", errors.New("PDF/A validation failed with an application error")
}
```

jraddaoui: Now seeing this, the second switch also seems unnecessary and could be solved with an `if`. Not a big deal, whatever you prefer, I just found the two nested switches with a single case a bit too much.

djjuhasz: I like the switch because it reads well to me and makes it easy to add more exit codes if we find that necessary in the future.  I can change it to and "if" conditional if you feel strongly about it.

verapdf_report.json
```diff
@@ -0,0 +1,385 @@
...
+                "successfulJobCount": 0
+            },
+            "repairSummary": {
+                "failedJobCount": 0,
+                "totalJobCount": 0,
+                "successfulJobCount": 0
+            }
+        }
+    }
+}
```
jraddaoui, line 385: @djjuhasz, I thought I added a comment about this file. Is it needed?

djjuhasz: Oops, no.  I'll delete it.

PR #99, opened 2024-12-06
WIP: Relate preprocessing and AIS workflows

jraddaoui:
Related to https://github.com/artefactual-sdps/enduro/pull/1087 and https://github.com/artefactual-sdps/enduro/issues/886.

- Generate a UUID in the preprocessing workflow and add it to the result.
- Receive preprocessing UUID in AIS workflow.
- TODO:
  - Preprocessing: send metadata file to AIS bucket.
  - AIS: fetch metadata file from AIS bucket.
  - AIS: remove METS parsing code.
  - AIS: remove metadata file download from AMSS.
  - AIS: remove metadata file from AIS bucket when the bundle has been stored.

### DIFF
--- a/internal/workflow/uuid.go
+++ b/internal/workflow/uuid.go
@@ -1,0 +1,14 @@
+package workflow
+
+import (
+	"github.com/google/uuid"
+	temporalsdk_workflow "go.temporal.io/sdk/workflow"
+)
+
+func GenUUID(ctx temporalsdk_workflow.Context) (uuid.UUID, error) {
+	var id uuid.UUID
+	gen := func(ctx temporalsdk_workflow.Context) interface{} { return uuid.New() }
+	enc := temporalsdk_workflow.SideEffect(ctx, gen)
+	err := enc.Get(&id)
+	return id, err
+}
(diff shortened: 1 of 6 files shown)


internal/workflow/uuid.go
```diff
@@ -0,0 +1,14 @@
...
+
+import (
+	"github.com/google/uuid"
+	temporalsdk_workflow "go.temporal.io/sdk/workflow"
+)
+
+func GenUUID(ctx temporalsdk_workflow.Context) (uuid.UUID, error) {
+	var id uuid.UUID
+	gen := func(ctx temporalsdk_workflow.Context) interface{} { return uuid.New() }
+	enc := temporalsdk_workflow.SideEffect(ctx, gen)
```
jraddaoui, line 11: Sadly, https://github.com/temporalio/sdk-go/issues/916. That's why I went with an UUID generator.

djjuhasz: Why not use the [WorkflowID](https://github.com/artefactual-sdps/enduro/blob/main/internal/package_/workflow.go#L30) from the processing Workflow?  You'd have to pass it to `preprocessing` and `postprocessing` too, but it's a unique identifier that is tied to the Workflow execution, and you can set it in test as well.

jraddaoui: Thanks @djjuhasz, that's a great point. And, now that you say that, we don't even need to pass that ID as it can be obtained from the context:

```go
parentID := temporalsdk_workflow.GetInfo(ctx).ParentWorkflowExecution.ID
```

I'll close this PR and do the remaining work in a new one.

Issue #98, opened 2024-12-04
Problem: migration packages will now contain a logical metadata file


sallain:
**Describe the bug**

The migration packages will include an extra metadata file going forward (the logical metadata extracted from Preservica). SFA would like there to be a check to confirm that the file has been included in the package, that it is valid, and then move it to the metadata directory. This affects the **BornDigitalAIP** and **DigitizedAIP** transfer types.

**Expected behavior**

Some activity (could re-use the _Validate SIP structure_ activity) will look for a file located at `package-name/data/additional/package-name-premis.xml`. The activity will confirm that the file is present and that the name of the file preceding `-premis.xml` matches the name of the package. 

Another activity (perhaps _Validate SIP metadata_ or _Validate PREMIS_) will validate the above listed file. It should be valid PREMIS. The output for the activity should indicate that this PREMIS file has been validated.

Finally, the file should be moved to the metadata directory during the _Restructure SIP_ activity.

I will provide sample packages once the structure is confirmed by SFA.

**Additional context**

See #97 for more information about the new package structure.

sallain:
LGTM! I see the error reported in two places:

![Image](https://github.com/user-attachments/assets/f47999a0-4fd5-4756-b962-fb2887fac144)

This is good, since SFA wants to have all preprocessing errors reported at once.

Issue #97, opened 2024-12-04
Problem: package structure for migrated AIPs has changed


sallain:
**Describe the bug**

The package structure for migration packages has changed. This affects the **BornDigitalAIP** and **DigitizedAIP** transfer types.

AIPs for both of these transfer types will be bagged, so there will be a second-level _/data_ directory that pushes everything else down in the hierarchy. Various preprocessing activities may fail with a bagged package, including:

- Identify SIP structure
- Validate SIP structure
- Verify SIP manifest
- Verify SIP checksums
- Validate SIP metadata
- Restructure SIP

Note that this may not be an exhaustive list!

Going forward, DigitizedAIPs will have the following structure:

```
7537ab2c-4e6b-4820-95bf-bd2c577351c3
├── bag-info.txt
├── bagit.txt
├── data
│   ├── additional
│   │   ├── 7537ab2c-4e6b-4820-95bf-bd2c577351c3-premis.xml
│   │   └── UpdatedAreldaMetadata.xml
│   └── content
│       ├── content
│       │   └── d_0000001
│       │       ├── 00000001.jp2
│       │       ├── 00000001_PREMIS.xml
...
│       │       ├── 00000020.jp2
│       │       ├── 00000020_PREMIS.xml
│       │       └── Prozess_Digitalisierung_PREMIS.xml
│       └── header
│           ├── old
│           │   └── SIP
│           │       └── metadata.xml
│           └── xsd
│               ├── ablieferung.xsd
│               ├── archivischeNotiz.xsd
│               ├── archivischerVorgang.xsd
│               ├── arelda.xsd
│               ├── base.xsd
│               ├── datei.xsd
│               ├── dokument.xsd
│               ├── dossier.xsd
│               ├── ordner.xsd
│               ├── ordnungssystemposition.xsd
│               ├── ordnungssystem.xsd
│               ├── paket.xsd
│               ├── provenienz.xsd
│               └── zusatzDaten.xsd
├── manifest-sha256.txt
├── manifest-sha512.txt
├── tagmanifest-sha256.txt
└── tagmanifest-sha512.txt
```

BornDigitalAIPs will have the following structure:

```
Test-AIP-Files/
├── bag-info.txt
├── bagit.txt
├── data
│   ├── additional
│   │   ├── Test-AIP-Files-premis.xml
│   │   └── UpdatedAreldaMetadata.xml
│   └── content
│       ├── content
│       │   ├── d0001
[...]

PR #100, opened 2024-12-12
Unbag bagged SIPs, fixes #94

### LINKED ISSUE #94: Feature: if a package is a bag, do not double-bag at the end of the child workflow

Currently, all packages are bagged at the end of the child workflow, before being sent to preservation. So far this hasn't been an issue as all of our testing material was unbagged. However, we have recommended to SFA that they bag all of the migration packages prior to ingest. In the current workflow, the package will be double-bagged. This seems inadvisable.

Check if the package is a bag, and if so then avoid double-bagging. I can think of two options:

1. [Update bag manifests](https://github.com/LibraryOfCongress/bagit-python/tree/master?tab=readme-ov-file#update-bag-manifests) -- this might be sufficient? I haven't tested it.

2. If option 1 is not sufficient, unbag the package at some point in the process (after #93) and then re-bag it at the end.

**Additional context**

djjuhasz:
Unbag bagged SIPs after validation to avoid issues with identification and double bagging.

### DIFF
--- a/internal/workflow/preprocessing.go
+++ b/internal/workflow/preprocessing.go
@@ -17,6 +17,7 @@ import (
 	"github.com/artefactual-sdps/preprocessing-sfa/internal/activities"
 	"github.com/artefactual-sdps/preprocessing-sfa/internal/enums"
 	"github.com/artefactual-sdps/preprocessing-sfa/internal/eventlog"
+	"github.com/artefactual-sdps/preprocessing-sfa/internal/localact"
 	"github.com/artefactual-sdps/preprocessing-sfa/internal/premis"
 	"github.com/artefactual-sdps/preprocessing-sfa/internal/sip"
 )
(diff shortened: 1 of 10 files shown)


internal/workflow/preprocessing.go
```diff
@@ -110,6 +111,29 @@ func (w *PreprocessingWorkflow) Execute(
 
 	localPath := filepath.Join(w.sharedPath, filepath.Clean(params.RelativePath))
 
+	// Unbag the SIP if it's a BagIt bag.
+	if fsutil.FileExists(filepath.Join(localPath, "bagit.txt")) {
```
sevein, line 115: It might be a good idea to wrap `fsutil.FileExists` with the SideEffect API or implement it as a local activity given that it relies on the local disk which is an unreliable resource. For the reader unclear about this problem:
* https://docs.temporal.io/workflows#intrinsic-non-deterministic-logic
* https://learn.temporal.io/tutorials/go/background-check/durable-execution/#intrinsic-non-deterministic-logic
* https://community.temporal.io/t/workflow-determinism/4027#intrinsic-4

workflowcheck doesn't seem able to identify this issue, but I think that's true in many other cases - they only seem to be paying attention to a subset of issues.
I haven't been actively following the work in this repo for a while, but I came across this pull request by chance while reviewing #97 and #98 yesterday - though I’m no longer looking at them, thanks!

djjuhasz: Hmm, good catch @sevein, I didn't have file I/O on my radar for non-deterministic code. :(  The Unbag activity itself is a no-op when the SIP is not a bag, but I'd rather not create and unbagging event if we aren't unbagging anything so I'll have to figure out how to conditionally create the event.

PR #101, opened 2024-12-17
Add logical metadata file validation, fixes #98

### LINKED ISSUE #98: Problem: migration packages will now contain a logical metadata file

djjuhasz:
- Add the logical metadata file to the expected SIP structure for AIP type SIPs
- Validate the logical metadata file against the PREMIS v3 XSD
- Move the logical metadata file to the "metadata/" directory when transforming the SIP for delivery to preservation


internal/activities/validate_premis_test.go
```diff
@@ -0,0 +1,154 @@
...
+					fs.WithFile("premis.xml", premisXML),
+				).Join("premis.xml"),
+			},
+			want: activities.ValidatePREMISResult{
+				Failures: []string{"premis.xml does not match expected metadata requirements"},
+			},
+		},
+		{
+			name:      "Returns a file not found failure",
+			validator: newFakeValidator().WithErr(errors.New("file not found")),
```
mcantelon, line 97: What are the pros and cons of having a mock object return a file not found error versus testing with a non-existent file?

internal/activities/validate_premis_test.go
```diff
@@ -0,0 +1,154 @@
...
+				if err == nil {
+					t.Errorf("error is nil, expecting: %q", tt.wantErr)
+				} else {
+					assert.ErrorContains(t, err, tt.wantErr)
+				}
+
+				return
+			}
+
+			t.Log(logbuf.String()) // Echo log for debugging.
```
mcantelon, line 146: Slick!

internal/activities/validate_structure.go
```diff
@@ -75,14 +82,24 @@ func (a *ValidateStructure) Execute(
...
+	if params.SIP.Type == enums.SIPTypeDigitizedSIP || params.SIP.Type == enums.SIPTypeDigitizedAIP && hasContentDir {
 		entries, err := os.ReadDir(params.SIP.ContentPath)
 		if err != nil {
 			return nil, fmt.Errorf("ValidateStructure: check for unexpected dossiers: %v", err)
 		}
 
-		if len(entries) > 1 {
+		dirs := 0
+		for _, e := range entries {
+			if e.IsDir() {
```
mcantelon, line 94: Good catch!

internal/premis/xsdv3.go
```diff
@@ -0,0 +1,6 @@
+package premis
+
+import _ "embed"
+
+//go:embed premis-v3-0.xsd
```
mcantelon, line 5: Very handy feature, this embed magic!

internal/sip/sip.go
```diff
@@ -114,3 +118,11 @@ func (s SIP) bornDigitalSIP() SIP {
 func (s SIP) Name() string {
 	return filepath.Base(s.Path)
 }
+
+func (s SIP) IsAIP() bool {
```
mcantelon, line 122: Good thinking moving these here!

mcantelon (approved):
This seems great! I had one question (for my own curiosity), but see any issues... the embedding of the PREMIS XSD is slick.

internal/activities/validate_premis_test.go
djjuhasz: @mcantelon I think the main pro of using a mock validator is that the tests are not dependent on xmllint being installed in the test environment. Unfortunately I negated that advantage by using the real "xmllint" validator in the first test, and sure enough that test failed in CI until I added xmllint to the CI runner. :thinking: The mock validator is also faster (though I'm sure the difference isn't noticeable in these small test cases), and allows more control over the return values. 

I think the main con of using a mock validator is if there is some integration issue between the `ValidatePREMIS` activity and `xmlvalidate.XMLLintValidator` then it won't be caught by these unit tests.  Ideally integration issues would be caught by integration tests, but we don't have any integration tests for this project. :(

internal/activities/validate_premis_test.go
djjuhasz: Yeah! It took me a while to figure out how to capture and test the log messages. :sweat_smile:

internal/activities/validate_premis_test.go
mcantelon: Good points! Yeah, integration tests definitely seem like they'd be handy (but also potentially a lot of work to create/maintain at this stage).

internal/sip/sip.go
djjuhasz: Yeah, it saves some repeat code. :)

internal/activities/validate_premis_test.go
djjuhasz: I've decided to remove the xmllint dependency in the first test here.  The deciding factor for me is that the tests should not fail if a developer runs them locally without xmllint installed.

Issue #93, opened 2024-11-28
Feature: validate incoming bag


sallain:
**Is your feature request related to a problem? Please describe.**

SFA's migration packages will arrive in the system as bags. In order to ensure that nothing has been corrupted or lost in transit, the bags should be validated.

**Describe the solution you'd like**

Add an activity that checks to see if the ingest is a bag and then validates the bag.

**Describe alternatives you've considered**

We could simply not validate the bag, but it seems like good practice to take advantage of the opportunity.

**Additional context**

mcantelon:
PR for CR (of bag validation activity... integration into workflow will be another PR): https://github.com/artefactual-sdps/temporal-activities/pull/38

fiver-watson:
tested by @sallain 2025-01-08 and failing validation

![Image](https://github.com/user-attachments/assets/cd86b50f-8585-4b6e-a787-f25e6623e771)

sallain:
LGTM! Bags are checked, marked as valid, and the child workflow proceeds.

PR #102, opened 2024-12-18
Add Bagit validation (#93)

mcantelon:
Added validation of Bag before it's unbagged.

### DIFF
--- a/internal/workflow/preprocessing_test.go
+++ b/internal/workflow/preprocessing_test.go
@@ -9,6 +9,7 @@ import (
 	"time"
 
 	"github.com/artefactual-sdps/temporal-activities/bagcreate"
+	"github.com/artefactual-sdps/temporal-activities/bagvalidate"
 	"github.com/artefactual-sdps/temporal-activities/ffvalidate"
 	"github.com/artefactual-sdps/temporal-activities/xmlvalidate"
 	"github.com/stretchr/testify/mock"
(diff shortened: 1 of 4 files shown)


internal/workflow/preprocessing_test.go
```diff
@@ -135,6 +137,14 @@ func (s *PreprocessingTestSuite) SetupTest(cfg *config.Configuration) {
 	s.sipPath = sp
 
 	// Register activities.
+	validator, err := bagit_gython.NewBagIt()
+	if err != nil {
+		s.T().Fatalf("initialize bagit: %v", err)
+	}
+	s.env.RegisterActivityWithOptions(
+		bagvalidate.New(validator).Execute,
```
djjuhasz, line 145: You can just use a nil validator here (e.g. `bagvalidate.New(nil).Execute`) and it won't make any difference to the results.  All of the activity calls in these workflow tests are mocked, so the activity is never actually run and the `validator.Validate()` method is never called.

djjuhasz (changes requested):
Looks good @mcantelon, just one change request.

mcantelon:
I've made that change, @djjuhasz ! Let me know if anything else needs tweaking!

internal/workflow/preprocessing_test.go
djjuhasz: @mcantelon you don't even need to use the `bagvalidate.NewNoopValidator()` I've discovered. You can literally just pass a `nil` for the validator:

```go
bagvalidate.New(nil).Execute,
```

djjuhasz (changes requested):

mcantelon:
Ah sweet! Fixed! Thanks @djjuhasz !

djjuhasz (approved):
Looking good! :shipit:

PR #103, opened 2025-01-09
Register bagvalidate activity

jraddaoui:
Create/clean bagit-gython validator and register bagvalidate activity in the preprocessing worker command.

Refs #93.

**Important!**

**This should make Bag validation work in systems using glibc but it won't work in the Alpine based image used locally. I'll address this issue in a different PR, for now Bags will fail on validation, other SIP formats should work as expected locally too.**


mcantelon (approved):
Looks good to me!

PR #105, opened 2025-01-13
Switch to a debian-slim base image

jraddaoui:
The `bagit-gython` library requires a system based on `glibc` while
Alpine uses `musl`. Change installation of JRE as the one from the
veraPDF CLI image is also based on `musl`.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -22,29 +22,22 @@ RUN --mount=type=cache,target=/go/pkg/mod \
 	-o /out/preprocessing-worker \
 	./cmd/worker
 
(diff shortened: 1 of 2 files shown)


jraddaoui:
I tried using a distroless image for consistency with Enduro, but we run veraPDF through a shell script. We could try a Java based distroless image, calling veraPDF with `java -jar ...` but I'm not sure if that will work with the current implementation.

djjuhasz (approved):
LGTM :+1:  Thanks @jraddaoui.

Issue #106, opened 2025-01-13
Feature: check to see if incoming SIP has already been ingested


sallain:
**Is your feature request related to a problem? Please describe.**

When processing a large volume of packages, it can be easy to upload the same package twice. This could also happen if package deposit is automated or in a variety of other situations. In any case, processing the same package twice is an unnecessary use of compute resources and storage space and should be avoided.

**Describe the solution you'd like**

I would like Enduro to check to see if the package has already been ingested. This could be done by computing a checksum for the package and recording this in a database, to be checked against in the future. All incoming packages will be compressed, so computing a checksum should be fast. Any checksum that is a repeat of one already recorded in the database will trigger a failure.

This should be optional :wink: 

**Describe alternatives you've considered**

This is a pretty high-level check that will not catch very similar packages. It would be possible to be much more specific - checking individual file checksums, for example, or checking certain metadata elements - but I think this will suffice for the migration (and therefore MVP). 

**Additional context**

jraddaoui:
An initial implementation for this issue has been included in the v0.7.0 release, it required three main changes:

**Receive compressed SIPs from Enduro:**

To be able to calculate a checksum for the entire SIP, they are now shared compressed between Enduro and the preprocessing workflow and the extraction happens in preprocessing.

**Add persistence layer:**

Include a persistence package for database interactions and define a small schema using Ent. For now, it only supports MySQL and creates a single table for the SIPs with their file name and calculated checksum.

**Check for duplicate SIP:**

The actual check is optional and can be enabled/disabled using the `checkDuplicates` config variable, when it's disabled the persistence layer/config is not needed and the DB is not considered.

When enabled, the check is the first thing that happens in the preprocessing workflow. It calculates the SIP checksum and tries to create and entry in the DB, if the checksum already exists it fails the workflow directly. This means that, if a SIP has already been ingested you won't be able to ingest it again unless the contents change.

---

Most of the expected errors should be content validation errors and would be okay to require different contents, but this doesn't consider unexpected or system errors (like we saw during testing, where the PIPs failed in AM due to a system error). Because this initial implementation is done in a custom preprocessing child workflow, the issue mentioned above doesn't have an "easy fix".

If it's needed, in the next iteration we could try to find a general and configurable solution in the parent Enduro workflow, or investigate options using the poststorage child workflows.

fiver-watson:
Included in latest release delivered to SFA on July 17th. Waiting to see if SFA has any feedback.

PR #107, opened 2025-01-14
WIP: Add persistence package and duplicates check

jraddaoui:
Refs #106.


jraddaoui:
This is ready for code review @djjuhasz. @mcantelon this could be a good PR to know a bit more about the persistence layer/package in other projects where we use Ent/Atlas.

@sevein I'd also like that you take a look when you have some time, specially about the `cgo` changes at build time to be able to use SQLite, I see other projects with support for it but not in the build process, I guess just for tests. If we see an issue with this approach we could just build as before and leave the SQLite side just for the tests.

hack/kube/overlays/dev/preprocessing-secret.yaml
```diff
@@ -9,6 +9,12 @@ stringData:
     verbosity = 2
 
     sharedPath = "/home/preprocessing/shared"
+    checkDuplicates = true
```
djjuhasz, line 12: Don't we want checkDuplicates to be false in the dev environment?

internal/config/config.go
```diff
@@ -35,6 +36,13 @@ type Configuration struct {
 	// Enduro and preservation processing.
 	SharedPath string
 
+	// CheckDuplicates enables or disables a check for SIPs that have already
+	// been processed. When enabled, the persistence configuration below will
+	// be required, and a SIP that has already been processed will fail the
+	// preprocessing workflow.
+	CheckDuplicates bool
+	Persistence     persistence.Config
```
djjuhasz, line 44: I think there should be a blank line after `CheckDuplicates` and no blank line after `Persistence`.

internal/persistence/db.go
```diff
@@ -0,0 +1,75 @@
...
+	_ "github.com/mattn/go-sqlite3"
+)
+
+// Open a database driver.
+func Open(driver, ds string) (db *sql.DB, err error) {
+	switch driver {
+	case "mysql":
+		db, err = OpenMySQL(ds)
+	case "sqlite3":
+		db, err = sql.Open(driver, ds)
```
djjuhasz, line 19: Why not put all this in an `OpenMySQL(ds)` function like with MySQL?

internal/persistence/db.go
```diff
@@ -0,0 +1,75 @@
...
+
+	default:
+		return nil, fmt.Errorf("database driver %q not supported", driver)
+	}
+
+	return db, err
+}
+
+// OpenMySQL opens the MySQL database driver.
+func OpenMySQL(ds string) (*sql.DB, error) {
```
djjuhasz, line 51: Does this need to be exported?

djjuhasz (approved):
Looks good to me @jraddaoui. :+1:  

I guess adding a database was inevitable, but it was nice to avoid it as long as we did. :wink:

hack/kube/overlays/dev/preprocessing-secret.yaml
jraddaoui: Sure, thanks!

sevein (approved):
You may want to consider [modernc.org/sqlite](https://pkg.go.dev/modernc.org/sqlite) instead. It's a CGO-free port used by many projects like Temporal, built with a C to Go translator (https://gitlab.com/cznic/ccgo).

jraddaoui:
Thank @sevein, I gave it a try and it requires some tweaks to make it work with Ent (diver `sqlite` vs `sqlite3`). We discussed yesterday that we won't be using SQLite in their environment, so I'll leave it as it is in `sdps_preprocessing` and I'll give it another try alongside the versioned migrations.

jraddaoui:
Thanks @djjuhasz, I think I've addressed all your feedback.

PR #110, opened 2025-01-15
Check for duplicated SIPs

jraddaoui:
In the preprocessing workflow, expect a compressed SIP instead of a directory. If checkDuplicates is enabled in the configuration, calculate the SIP checksum and persist the SIP name and checksum in the DB. If there is already an entry with the same checksum, fail the workflow with a validation error.

- Add activity to calculate the SIP checksum.
- Add local activity to check for duplicates.
- Include archiveextract activity from temporal-activities.
- Add events for the three activities to the result.

Refs #106, follow up to #107.


cmd/worker/workercmd/cmd.go
```diff
@@ -75,11 +77,47 @@ func (m *Main) Run(ctx context.Context) error {
...
+		)
+		if m.cfg.Persistence.Migrate {
+			err = m.dbClient.Schema.Create(ctx)
+			if err != nil {
+				m.logger.Error(err, "Error migrating database.")
+				return err
+			}
+		}
+		ps = entclient.New(m.dbClient)
+	}
```
jraddaoui, line 106: Just moved the previous code from below and added the `ps` variable declaration and assignment.

cmd/worker/workercmd/cmd.go
```diff
@@ -75,11 +77,47 @@ func (m *Main) Run(ctx context.Context) error {
...
 	w.RegisterWorkflowWithOptions(
-		workflow.NewPreprocessingWorkflow(m.cfg.SharedPath).Execute,
+		workflow.NewPreprocessingWorkflow(m.cfg.SharedPath, m.cfg.CheckDuplicates, ps).Execute,
 		temporalsdk_workflow.RegisterOptions{Name: m.cfg.Temporal.WorkflowName},
 	)
 
+	w.RegisterActivityWithOptions(
+		activities.NewChecksumSIP().Execute,
+		temporalsdk_activity.RegisterOptions{Name: activities.ChecksumSIPName},
+	)
```
jraddaoui, line 116: It won't be executed if `checkDuplicates` is set to `false`, but I thought it doesn't hurt to register it in any case. Not sure if that's really a good idea, I could move the registration to the if block above.

internal/activities/checksum_sip.go
```diff
@@ -0,0 +1,41 @@
...
+}
+
+func (a *ChecksumSIP) Execute(ctx context.Context, params *ChecksumSIPParams) (*ChecksumSIPResult, error) {
+	f, err := os.Open(params.Path)
+	if err != nil {
+		return nil, fmt.Errorf("ChecksumSIP: open SIP: %v", err)
+	}
+	defer f.Close()
+
+	h := sha256.New()
```
jraddaoui, line 35: Less collision probability with a small cost in performance compared to MD5.

internal/workflow/preprocessing.go
```diff
@@ -112,6 +118,64 @@ func (w *PreprocessingWorkflow) Execute(
 
 	localPath := filepath.Join(w.sharedPath, filepath.Clean(params.RelativePath))
 
+	if w.checkDuplicates {
+		// Calculate SIP checksum.
+		ev := result.newEvent(ctx, "Calculate SIP checksum")
```
jraddaoui, line 123: Help with event names and outcomes would be appreciated, I got lost between "duplicate/duplicates/duplicated SIP" and went with basic event names and the same name plus passed/failed in the outcome.

internal/workflow/preprocessing.go
```diff
@@ -112,6 +118,64 @@ func (w *PreprocessingWorkflow) Execute(
...
+		if e != nil {
+			result.systemError(ctx, e, ev, "Check duplicate SIP failed")
+			return result, nil
+		}
+		if checkDuplicate.Duplicated {
+			result.validationError(
+				ctx,
+				ev,
+				"Check duplicate SIP failed",
+				[]string{"Found a SIP with the same checksum"},
```
djjuhasz, line 157: Suggested: "A previously submitted SIP has the same checksum".

internal/workflow/preprocessing.go
```diff
@@ -112,6 +118,64 @@ func (w *PreprocessingWorkflow) Execute(
...
+		).Get(ctx, &checkDuplicate)
+		if e != nil {
+			result.systemError(ctx, e, ev, "Check duplicate SIP failed")
+			return result, nil
+		}
+		if checkDuplicate.Duplicated {
+			result.validationError(
+				ctx,
+				ev,
+				"Check duplicate SIP failed",
```
djjuhasz, line 156: Suggested: "SIP is a duplicate". I know this doesn't match the format of the other validation messages, but I think it's important to distinguish between the "check duplicate" activity failing, and rejecting the SIP because it is a duplicate.

internal/workflow/preprocessing.go
```diff
@@ -112,6 +118,64 @@ func (w *PreprocessingWorkflow) Execute(
...
+				[]string{"Found a SIP with the same checksum"},
+			)
+			return result, nil
+		}
+		ev.Succeed(ctx, "Check duplicate SIP passed")
```
djjuhasz, line 161: Suggested: "SIP does not match any previous SIPs".

internal/workflow/preprocessing.go
```diff
@@ -112,6 +118,64 @@ func (w *PreprocessingWorkflow) Execute(
...
+		withFilesysActOpts(ctx),
+		archiveextract.Name,
+		&archiveextract.Params{SourcePath: localPath},
+	).Get(ctx, &archiveExtract)
+	if err != nil {
+		result.systemError(ctx, e, ev, "Extract SIP failed")
+		return result, nil
+	}
+	localPath = archiveExtract.ExtractPath
+	ev.Succeed(ctx, "Extract SIP passed")
```
djjuhasz, line 177: Suggested: "SIP extracted".

djjuhasz (approved):
This looks good @jraddaoui, I mostly provided suggestions for the activity messages.

jraddaoui:
Thanks a lot @djjuhasz!

PR #111, opened 2025-01-16
Fix relative path after SIP extraction

jraddaoui:
Refs #106.

### DIFF
--- a/internal/workflow/preprocessing.go
+++ b/internal/workflow/preprocessing.go
@@ -168,16 +168,21 @@ func (w *PreprocessingWorkflow) Execute(
 	// Extract SIP.
 	ev := result.newEvent(ctx, "Extract SIP")
 	var archiveExtract archiveextract.Result
-	err := temporalsdk_workflow.ExecuteActivity(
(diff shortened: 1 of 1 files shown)


internal/workflow/preprocessing.go
```diff
@@ -221,14 +226,13 @@ func (w *PreprocessingWorkflow) Execute(
...
-		rp, err := filepath.Rel(w.sharedPath, localPath)
-		if err != nil {
-			return nil, fmt.Errorf("Invalid path")
+		result.RelativePath, e = filepath.Rel(w.sharedPath, localPath)
+		if e != nil {
+			result.systemError(ctx, e, ev, "Unbagging the SIP has failed")
+			return result, nil
 		}
-		result.RelativePath = rp
+		ev.Succeed(ctx, "SIP unbagged")
```
jraddaoui, line 235: I was testing only Bags so this relative path update was hiding the error. I've changed this too, so it reports the possible error in the related event and the result.

djjuhasz (approved):
LGTM :+1:

PR #115, opened 2025-01-19
Fail structure validation if any directory is empty

mcantelon:
Added logic to the validate SIP structure activity to detect empty directories and fail validation if they exist.

### DIFF
--- a/internal/activities/validate_structure.go
+++ b/internal/activities/validate_structure.go
@@ -3,6 +3,7 @@ package activities
 import (
 	"context"
 	"fmt"
+	"io/fs"
 	"os"
 	"path/filepath"
 	"slices"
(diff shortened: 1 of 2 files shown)


internal/activities/validate_structure.go
```diff
@@ -34,6 +36,44 @@ func (a *ValidateStructure) Execute(
...
+	paths := make(map[string]int)
+
+	err := filepath.WalkDir(params.SIP.Path, func(path string, d fs.DirEntry, err error) error {
+		if err != nil {
+			return err
+		}
+
+		if path != params.SIP.Path {
+			// Assemble relative path.
+			relativePath := strings.Replace(path, params.SIP.Path+"/", "", 1)
```
jraddaoui, line 49: Why not using `filepath.Rel()`?

internal/activities/validate_structure.go
```diff
@@ -34,6 +36,44 @@ func (a *ValidateStructure) Execute(
...
+		return nil
+	})
+	if err != nil {
+		return nil, fmt.Errorf("ValidateStructure: check for empty directories: %v", err)
+	}
+
+	// Report any empty subdirectories.
+	for path, children := range paths {
+		if children == 0 {
+			failures = append(failures, fmt.Sprintf("SIP subdirectory '%s' is empty", path))
```
jraddaoui, line 69: Instead of `'%s'` , `%q` will quote the string.

jraddaoui (approved):
LGTM, thanks @mcantelon!

mcantelon:
Thanks @jraddaoui !

PR #116, opened 2025-01-22
Detect invalid characters in file/directory names

mcantelon:
Added logic to validate structure to return validation errors if a file or directory's name contains characters incompatible with Archivematica.

### DIFF
--- a/internal/activities/validate_structure.go
+++ b/internal/activities/validate_structure.go
@@ -7,6 +7,7 @@ import (
 	"os"
 	"path/filepath"
 	"slices"
+	"strings"
 
 	"github.com/artefactual-sdps/preprocessing-sfa/internal/enums"
 	"github.com/artefactual-sdps/preprocessing-sfa/internal/fsutil"
@@ -35,7 +36,7 @@ func (a *ValidateStructure) Execute(
 ) (*ValidateStructureResult, error) {
 	var failures []string
 
-	// Check for empty directories.
+	// Check for empty directories and invalid (Archivematica incompatible) file/directory names.
 	paths := make(map[string]int)
 
 	err := filepath.WalkDir(params.SIP.Path, func(path string, d fs.DirEntry, err error) error {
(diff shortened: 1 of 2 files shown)


internal/activities/validate_structure_test.go
```diff
@@ -180,6 +196,13 @@ func TestValidateStructure(t *testing.T) {
 				Failures: []string{"More than one dossier in the content directory"},
 			},
 		},
+		{
+			name:   "Returns a failure when a SIP a directory name with an invalid character",
```
jraddaoui, line 200: `directory/file`

internal/activities/validate_structure.go
```diff
@@ -171,3 +176,29 @@ func extraNodes(sipBase, path string, expected []string, matchDir bool) ([]strin
...
 	return extras, nil
 }
+
+func validateName(name string) bool {
+	validChars := "-_.()"
+
+	// Generate string containing valid characters.
+	validChars = appendRuneRange(validChars, 'a', 'z')
+	validChars = appendRuneRange(validChars, 'A', 'Z')
+	validChars = appendRuneRange(validChars, '0', '9')
```
jraddaoui, line 186: This is nice, but I think it would be a lot clearer with a long string constant. It will also prevent from calculating it each time.

jraddaoui (changes requested):
Thanks @mcantelon!

mcantelon:
Thanks @jraddaoui ! PR updated!

internal/activities/validate_structure.go
```diff
@@ -171,3 +176,16 @@ func extraNodes(sipBase, path string, expected []string, matchDir bool) ([]strin
 
 	return extras, nil
 }
+
+func validateName(name string) bool {
+	const validChars = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789-_.()"
+
+	// Make sure only valid characters exist in name.
```
jraddaoui, line 183: This type of comments are better above the function name:

```
// validateName makes sure only valid characters exist in name.
func validateName(name string) bool {
```

Not a big deal in this case for a small private function, but that way it documents the function and you can read it when you use it.

jraddaoui (approved):
Thanks @mcantelon!

mcantelon:
Thanks @jraddaoui !

Issue #122, opened 2025-02-10
Feature: check that package name complies with naming rules


sallain:
**Is your feature request related to a problem? Please describe.**

SFA has defined naming conventions for transfers. Transfers should be rejected if the name does not comply.

**Describe the solution you'd like**

Create an activity to check that the name of a package complies with the naming rules for that transfer type - e.g. a BornDigital SIP must be named with a particular pattern. The naming pattern can be fully defined or partially defined (e.g. must start with a certain prefix). There might be one naming pattern per transfer type or multiple naming patterns.

Note that this feature should only apply to the SIP transfer types; AIPs will have their own naming conventions, some of which might be historic/out of date, so I think we shouldn't worry about them.

Naming convention for BornDigitalSIP (but not BornDigitalAIP):

`SIP_ [delivery date YYYYMMDD]_[name of the delivering office]_[reference]`
eg. `SIP_20250101_someoffice_refno` or `SIP_20241016_someoffice`

The reference can be used optionally by the delivering office, so it should be possible to have `_[reference]` but not required.

Naming convention for DigitizedSIP (but NOT DigitizedAIP):

`SIP_[YYYYMMDD]_Vecteur_[Digitization order ID]`

The error message should indicate that the name is noncompliant, e.g. "The name of this package does not comply with the naming requirments."

Finally, a PREMIS validation event should be written to indicate that the name was checked.

**Describe alternatives you've considered**

None.

**Additional context**

mcantelon:
I've got a working PR roughed out, but had a could of questions:

What range of years should be valid for SIP names? 1800s, 1900s, 2000s? Any four digit year at all?

Does Case matter for "Vecteur" in names? Does it have to be as I had it in quotes or are "vecteur" and other variants also allowed?

sallain:
@mcantelon I've asked SFA and they would not like any restrictions on the date range - they sometimes use fake dates for older material. As long as it's a four-digit year that will be fine. 

Case doesn't matter for vecteur, it's case-insensitive.

mcantelon:
PR for CR: https://github.com/artefactual-sdps/preprocessing-sfa/pull/127

mcantelon:
PR merged!

fiver-watson:
This works in terms of validating SIP names. Nice work!

Internal reminder for ourselves that once SFA makes final decisions, we need to update this task so that it will properly  validate migration AIP package names as well. 

In terms of feedback: 

[...]

PR #127, opened 2025-02-23
Validate SIP names

mcantelon:


internal/activities/validate_sip_name.go
```diff
@@ -0,0 +1,38 @@
+package activities
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/artefactual-sdps/preprocessing-sfa/internal/sip"
+)
+
+const ValidateSIPNameName = "validate-transfer-name"
```
djjuhasz, line 10: Can the string be "validate-sip-name" too or is there a reason to keep using "transfer"?

djjuhasz: Haha, ValidateSIPNameName is pretty cute. 😆

internal/sip/sip_test.go
```diff
@@ -65,10 +53,14 @@ func TestNew(t *testing.T) {
...
-		fs.WithDir("header"),
-	)
+	bornDigitalSIPName := "SIP_20201201_someoffice_someref"
+	bornDigitalSIPTempDir := fs.NewDir(t, "",
+		fs.WithDir(bornDigitalSIPName,
+			fs.WithDir("content"),
+			fs.WithDir("header"),
+		),
+	).Path()
+	bornDigitalSIPPath := filepath.Join(bornDigitalSIPTempDir, bornDigitalSIPName)
```
djjuhasz, line 64: I think it would good to have a test case for a valid born digital SIP name with no reference (e.g. "SIP_20010106_someoffice" -- I'm not sure if the current RegEx will allow it.

internal/sip/sip.go
```diff
@@ -126,3 +127,20 @@ func (s SIP) IsAIP() bool {
 func (s SIP) IsSIP() bool {
 	return s.Type == enums.SIPTypeBornDigitalSIP || s.Type == enums.SIPTypeDigitizedSIP
 }
+
+func (s SIP) HasValidName() bool {
+	yyyymmdd := "(\\d{4})(\\d{2})(\\d{2})"
+	alphnum := "[a-zA-Z0-9]"
+
+	if s.Type == enums.SIPTypeBornDigitalSIP {
+		match, _ := regexp.MatchString(fmt.Sprintf("^SIP_%s_%s+[_]?%s+$", yyyymmdd, alphnum, alphnum), s.Name())
```
djjuhasz, line 136: I'm not clear from @sallain's description in #122 if the check should allow additional text after the matching pattern (e.g. "SIP_20250101_someoffice_refno_additional") or not. 🤔

internal/workflow/preprocessing.go
```diff
@@ -267,6 +267,24 @@ func (w *PreprocessingWorkflow) Execute(
...
+	if e != nil {
+		result.systemError(ctx, e, ev, "SIP name validation has failed")
+		return result, nil
+	}
+	if ValidateSIPName.Failures != nil {
+		result.validationError(ctx, ev, "SIP name validation has failed", ValidateSIPName.Failures)
+	} else {
+		ev.Succeed(ctx, "SIP name matches validation criteria")
+	}
+
```
djjuhasz, line 287: Is this writing a PREMIS event when validation succeeds?

djjuhasz (changes requested):
Hi @mcantelon. Looking pretty good, but I had a few questions about the exact validation criteria.  I think writing a PREMIS event is also missing, but you could add that in a subsequent PR.

internal/sip/sip.go
mcantelon: I'm going to assume not for now.

internal/workflow/preprocessing.go
mcantelon: Oops! I've added this now.

internal/sip/sip.go
```diff
@@ -126,3 +127,27 @@ func (s SIP) IsAIP() bool {
...
 }
+
+func (s SIP) HasValidName() bool {
+	yyyymmdd := "(\\d{4})(\\d{2})(\\d{2})"
+	alphaNum := "[a-zA-Z0-9]"
+	alphaNumAndUnderscore := "[a-zA-Z0-9_]"
+
+	if s.Type == enums.SIPTypeBornDigitalSIP {
+		match, _ := regexp.MatchString(
+			fmt.Sprintf("^SIP_%s_%s+[_]?%s+$", yyyymmdd, alphaNum, alphaNumAndUnderscore),
```
djjuhasz, line 138: I think the square brackets in the `[_]?` part of this regex are unnecessary, as `_?` would be equivalent. I also think that the final `%s+$` is misleading as it implies that the refno must match one or more times (it's matching names with no refno because the third underscore is optional which allows it to match on the final letters of the office).

I'd think the following regex would be more readable:

`fmt.Sprintf("^SIP_%s_%s+(_%s+)?$", yyyymmdd, alphaNum, alphaNumAndUnderscore),`

See this example: https://regex101.com/r/2G8G8k/1

internal/activities/validate_sip_name_test.go
```diff
@@ -0,0 +1,178 @@
...
+		params  activities.ValidateSIPNameParams
+		want    activities.ValidateSIPNameResult
+		wantErr string
+	}{
+		{
+			name:   "Validates the name of a born digital SIP",
+			params: activities.ValidateSIPNameParams{SIP: bornDigitalSIP},
+		},
+		{
+			name:   "Validates the name of a born digital SIP with a hyphen in the reference number",
```
djjuhasz, line 71: I think "hyphen" should be replaced with "underscore" here.

internal/activities/validate_sip_name_test.go
```diff
@@ -0,0 +1,178 @@
...
+					),
+				},
+			},
+		},
+		{
+			name:   "Validates the name of a digitized SIP",
+			params: activities.ValidateSIPNameParams{SIP: digitizedSIP},
+		},
+		{
+			name:   "Validates the name of a digitized SIP with a hyphen in the reference number",
```
djjuhasz, line 95: I think "hyphen" should be replaced with "underscore" here.

internal/activities/validate_sip_name_test.go
```diff
@@ -0,0 +1,178 @@
...
+	)
+	assert.NilError(t, err)
+
+	bornDigitalSIPBadName := "somref"
+	bornDigitalSIPBad, err := sip.New(
+		filepath.Join(bornDigitalSIPTempDir(t, bornDigitalSIPBadName), bornDigitalSIPBadName),
+	)
+	assert.NilError(t, err)
+
+	digitizedSIPName := "SIP_20010106_Vecteur_someref"
```
djjuhasz, line 42: It's not clear from the issue ticket, but I wonder if the DigitizedSIP format requirement of
`SIP_[YYYYMMDD]_Vecteur_[Digitization order ID]` implies that the "Digitization order ID" is a number, rather than an alphanumeric string?

I think you can leave it as is for now, and @sallain or the client can clarify the exact requirements.

djjuhasz (approved):
Looks good @mcantelon! I just added a couple of minor suggestions.

internal/sip/sip.go
mcantelon: Ah, nice! Good call!

internal/activities/validate_sip_name_test.go
mcantelon: Sounds good!

mcantelon:
Thanks @djjuhasz ! I've made those tweaks!

Issue #118, opened 2025-01-30
Problem: PREMIS event is not created for format validation


sallain:
**Describe the bug**

#63 added the ability to validate a file against the format specification using a given tool (VeraPDF in our initial implementation), but we forgot to generate a PREMIS event that records the validation.

**To Reproduce**

1. Transfer a SIP containing a valid PDF/A
2. Review the resulting premis.xml file in the AIP

**Expected behavior**

Create PREMIS events for each object validated. Here's a sample event:

```
  <premis:event>
    <premis:eventIdentifier>
      <premis:eventIdentifierType>UUID</premis:eventIdentifierType>
      <premis:eventIdentifierValue>9a38ee61-e1b1-46c7-ad82-e8692f07f9e5</premis:eventIdentifierValue>
    </premis:eventIdentifier>
    <premis:eventType>validation</premis:eventType>
    <premis:eventDateTime>2025-01-11T00:11:11Z</premis:eventDateTime>
    <premis:eventDetailInformation>
      <premis:eventDetail>name=&quot;Validate SIP file formats&quot;</premis:eventDetail>
    </premis:eventDetailInformation>
    <premis:eventOutcomeInformation>
      <premis:eventOutcome>valid</premis:eventOutcome>
      <premis:eventOutcomeDetail>
        <premis:eventOutcomeDetailNote>File format complies with specification.</premis:eventOutcomeDetailNote>
      </premis:eventOutcomeDetail>
    </premis:eventOutcomeInformation>
    <premis:linkingAgentIdentifier>
      <premis:linkingAgentIdentifierType valueURI="http://id.loc.gov/vocabulary/identifiers/local">url</premis:linkingAgentIdentifierType>
      <premis:linkingAgentIdentifierValue>https://github.com/artefactual-sdps/preprocessing-sfa</premis:linkingAgentIdentifierValue>
    </premis:linkingAgentIdentifier>
  </premis:event>
```

The agent should be the tool that performed the validation:

<premis:agent>
  <premis:agentIdentifier>
    <premis:agentIdentifierType valueURI="http://id.loc.gov/vocabulary/identifiers/local">url</premis:agentIdentifierType>
    <premis:agentIdentifierValue>https://verapdf.org</premis:agentIdentifierValue> 
  </premis:agentIdentifier>
  <premis:agentName>VeraPDF 1.26.2</premis:agentName>
  <premis:agentType>software</premis:agentType>
</premis:agent>

**Screenshots**

[...]

PR #128, opened 2025-02-25
Add PREMIS agent for veraPDF (#118)

mcantelon:


internal/workflow/preprocessing.go
```diff
@@ -550,20 +550,22 @@ func writePREMISFile(ctx temporalsdk_workflow.Context, sip sip.SIP) error {
 	}
 
 	// Add PREMIS events for validate file activity.
+	veraPDFAgent := premis.Agent{
+		Type:    "software",
+		Name:    "VeraPDF 1.26.2",
```
jraddaoui, line 555: I think we need to make this configurable or read the version from the system. For now, I'd add it to the `[filevalidate.verapdf]` configuration as `premisAgentName`:

https://github.com/artefactual-sdps/preprocessing-sfa/blob/main/hack/kube/overlays/dev/preprocessing-secret.yaml#L59-L60

jraddaoui (changes requested):
Thanks @mcantelon!

internal/workflow/preprocessing.go
jraddaoui: I see, we already had this hardcoded, so I'll merge as it is and make a note in the issue, thanks @mcantelon!

jraddaoui (approved):

PR #130, opened 2025-03-03
Get veraPDF version dynamically

mcantelon:
Execute the veraPDF command, when starting the worker, to get the current veraPDF version so it can be used when recording PREMIS events.

### DIFF
--- a/internal/fvalidate/verapdf_validator.go
+++ b/internal/fvalidate/verapdf_validator.go
@@ -4,6 +4,7 @@ import (
 	"errors"
 	"fmt"
 	"os/exec"
+	"strings"
 
 	"github.com/go-logr/logr"
 
@@ -81,4 +82,33 @@ func (v *veraPDFValidator) Validate(path string) (string, error) {
 		v.logger.Info("veraPDF validate", "exit code", e.ExitCode(), "STDERR", string(e.Stderr))
 		return "", errors.New("PDF/A validation failed with an application error")
 	}
+}
+
+func RunCommand(name string, args ...string) (string, error) {
+	result := exec.Command(name, args...) // #nosec: G204 -- trusted path.
+
+	output, err := result.Output()
+	if err != nil {
+		return "", err
+	}
+
+	return string(output), nil
+}
+
+var Run = RunCommand
+
+func (v *veraPDFValidator) Version() (string, error) {
+	// If the veraPDF cmd path is not set then skip returning the version.
+	if v.cmd == "" {
+		return "", nil
+	}
+
+	output, err := Run(v.cmd, "--version")
+	if err != nil {
+		return "", err
+	}
+
+	lines := strings.Split(output, "\n")
+
+	return lines[0], nil
 }
--- a/internal/fvalidate/verapdf_validator_test.go
+++ b/internal/fvalidate/verapdf_validator_test.go
@@ -2,6 +2,7 @@ package fvalidate_test
 
 import (
 	"fmt"
+	"strings"
 	"testing"
 
 	"github.com/go-logr/logr"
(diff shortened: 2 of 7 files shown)


mcantelon:
Thanks @jraddaoui ! I've made those changes. Seems like I could potentially get the version from the file validate activity as part of the results, rather than passing it as an argument when creating the preprocessing workflow, although it wouldn't be completely straightforward to do (as I'd have to cycle through the activity's validators property to find the veraPDF validator to get the version from).

internal/fvalidate/verapdf_validator.go
```diff
@@ -82,3 +83,16 @@ func (v *veraPDFValidator) Validate(path string) (string, error) {
 		return "", errors.New("PDF/A validation failed with an application error")
 	}
 }
+
+func (v *veraPDFValidator) Version() (string, error) {
+	result := exec.Command(v.cmd, "--version") // #nosec: G204 -- trusted path.
```
jraddaoui, line 88: Following what `Validate` does, we should check if `v.cmd` is set:

```
if v.cmd == "" {
    return "", nil
}
```

internal/fvalidate/verapdf_validator_test.go
```diff
@@ -94,3 +95,23 @@ func TestValidate(t *testing.T) {
...
+
+func TestVeraPDFVersionError(t *testing.T) {
+	t.Parallel()
+
+	v := fvalidate.NewVeraPDFValidator("false", logr.Discard())
+	version, err := v.Version()
+
+	assert.Error(t, err, "exit status 1")
+	assert.Equal(t, version, "")
+}
```
jraddaoui, line 117: Just realized about these test. How hard is to mock the command execution? I'd prefer something along those lines that actually running these commands.

jraddaoui (commented):
Thanks @mcantelon!

jraddaoui:
> Thanks @jraddaoui ! I've made those changes. Seems like I could potentially get the version from the file validate activity as part of the results, rather than passing it as an argument when creating the preprocessing workflow, although it wouldn't be completely straightforward to do (as I'd have to cycle through the activity's validators property to find the veraPDF validator to get the version from).

That would make sense, and prevent a possible issue where the activity is executed by another worker. Not a big deal right now as we require to share the filesystem with Enduro in these workers ATM.

internal/fvalidate/verapdf_validator_test.go
mcantelon: I was pondering that as well... I'll look into mocking as that should be more consistent.

mcantelon: I've added mocking of the command execution.

mcantelon:
@jraddaoui I've implemented the feedback (leaving  for later getting the version from the file validate activity rather than passing it as an argument). Thanks for all the reviewing!

internal/fvalidate/verapdf_validator.go
```diff
@@ -82,3 +83,32 @@ func (v *veraPDFValidator) Validate(path string) (string, error) {
...
+
+	output, err := result.Output()
+	if err != nil {
+		return "", err
+	}
+
+	return string(output), nil
+}
+
+var Run = RunCommand
```
jraddaoui, line 98: Ideally, this would be a dependency injected when the validator is created, to avoid replacing it in tests. No need to do this now, but it would be nice to create an interface that accepts an `exec.Command` or a mocked version, so you can pass that directly and not have a default function.

internal/fvalidate/verapdf_validator.go
```diff
@@ -82,3 +83,32 @@ func (v *veraPDFValidator) Validate(path string) (string, error) {
...
+
+func (v *veraPDFValidator) Version() (string, error) {
```
jraddaoui, line 100: I think it would be good to add this to the validator interface (and its mock):

- Add it here: https://github.com/artefactual-sdps/preprocessing-sfa/blob/main/internal/fvalidate/fvalidate.go#L10
- Run `make gen-mock` to update the fake implementation

internal/fvalidate/verapdf_validator.go
```diff
@@ -82,3 +83,32 @@ func (v *veraPDFValidator) Validate(path string) (string, error) {
 		return "", errors.New("PDF/A validation failed with an application error")
 	}
 }
+
+func RunCommand(name string, args ...string) (string, error) {
+	result := exec.Command(name, "--version") // #nosec: G204 -- trusted path.
```
jraddaoui, line 88: Use `args` instead of `"--version"`.

jraddaoui (changes requested):
Thanks @mcantelon! Don't worry about making it a dependency for now, but it would be nice to address the other two comments.

jraddaoui (approved):
Thanks Mike! Add `[skip-codecov]` to the squashed commit before merging ;)

mcantelon:
Thanks @jraddaoui !

PR #132, opened 2025-03-13
Increase timeouts for GetAIPPath activity

### LINKED ISSUE #126: Problem: timeout for GetAIPPathActivity is too short
**Describe the bug**

Per @djjuhasz (lightly edited):

The total time allowed for the "GetAIPPathActivity" to run is 5 seconds. The "GetAIPPathActivity" is sending a request to the AMSS API to get the AIP path, so if the AMSS was unreachable or is just busy and can't respond within 5 seconds, the workflow fails.  Note also that this is a "ScheduleToCloseTimeout" which means any retries would also need to complete within that 5 second window.

**Expected behavior**

Increase the "ScheduleToCloseTimeout" to a much larger value (maybe 10 minutes?).  The activity currently has a retry policy of making three retries, with an initial interval of one second, and a backoff multiplier of two (e.g. second attempt will be 1 second after the first, the third attempt will be 2 seconds after that).  Because Archviematica can run for a long time we should also bump up the initial retry interval to 30 seconds to allow more time for the server load to go down.

djjuhasz:
Fixes #126

- Make GetAIPPath a regular Temporal activity
- Increase schedule-to-close timeout to 10 minutes
- Increase initial retry interval to 15 seconds
- Increase maximum retries to 5
- Move AMSS client to it's own package and define a service interface
- Inject the AMSS client into the AIS workflow in the command runner instead of when the workflow is created
- Add unit tests for the GetAIPPath Activity


internal/amss/service.go
```diff
@@ -0,0 +1,11 @@
+package amss
+
+import (
+	"context"
+	"io"
+)
+
+type Service interface {
```
jraddaoui, line 8: Why calling it `Service`? And then `Client` and `NewClient`? We have called them `Client` and `clientImpl` in other interfaces/implementations.

jraddaoui (changes requested):
LGTM! Just a couple of minor details.

internal/amss/service.go
djjuhasz: I've gone with the name "pooledClient" for your AMSS client implementation because it's based on the hashicorp/cleanhttp pooled HTTP client.

jraddaoui (approved):
Thanks @djjuhasz, looks great to me!

PR #134, opened 2025-03-14
Remove duplicate AIS activity registration

djjuhasz:
Use the AIS `RegisterActivities()` function to register activities in the AIS workflow code and tests.


jraddaoui (approved):
Thanks!

PR #136, opened 2025-03-24
Don't add inapplicable veraPDF PREMIS events (#122)

mcantelon:
File validators only check certain types of files. Added activity to only add PREMIS events for files that have actually been checked by a validator.


jraddaoui (approved):
LGTM, thanks Mike!

We should start thinking about updating the PREMIS file from within the activities generating the events, it seems really inefficient to identify all files multiple times across the workflow.

mcantelon:
Thanks @jraddaoui !

Yeah, good point!

Issue #43, opened 2024-08-14
Epic: Content validation messages are not always clear to operators


fiver-watson:
**Describe the bug**

Sprint 2 Talk-aloud testing found a number of issues when performing usability tests related to packages that fail validation. Problems observed include: 

* Users did not know where to find their package
* Users stopped reading after encountering long absolute paths
* Error messages were often too vague for users to understand what was required to resolve the issue

After the talk-aloud testing, DG prepared an initial proposal based on the feedback received on how to revise the content validation error messages to make them more useful to operators. Two different groups at SFA reviewed these proposals and provided additional feedback. DG has now incorporated some of this additional feedback into a revised proposal for improving the error messages. 

**Proposed solution** 

See the attached PDF. A copy of this can also be found in the internal team Drive, [here](https://docs.google.com/document/d/1F5pU2pQ18QQ2NZP23PHAIi_oxGmhPE0YwJAQDwgBWxI/edit)

-----

**UPDATE 2024-10-02**

We got some additional feedback during the SFA onsite workshops, specifically about the proposed "Validate Metadata" error messages. See the notes [here](https://docs.google.com/document/d/1gRDpraDLr8zziLKCb_9jU7bkTIfMI617D_dGgMUQiVE/edit#heading=h.afgftlv701dn). Quick summary: 

* Providing a line number in the metadata.xml is not helpful
* They will be using PackageHandler to fix things, not opening the file directly
* As such, providing info on the actual XML element that was a problem is more useful
* Radda looked at the XSD output and we should be able to get the specific element
* If getting the element is not possible THEN we can fall back to the line number
   * In such a case, PackageHandler probably won't be able to parse either, so line number is helpful for manual inspection 

Also: while the proposal mockups include hyperlinks to either the failed packages directly, or the related failed minIO bucket, we noted during the review that this was a nice-to-have, but not a requirement if too complex. 

-----

**UPDATE 2025-02-18**

Sara and I have reviewed all current Enduro task success and error messages in light of ongoing SFA discussions and other related work. An updated spreadsheet with proposed changes is here: 

* https://docs.google.com/spreadsheets/d/1gqRsT3-g9MFv8yWDKUp6UjX2aw9aBYfOqRZHQUX6Lnc/edit?gid=1946880320#gid=1946880320

fiver-watson:
[...]

PR #138, opened 2025-03-27
Improve content and system error messages

djjuhasz:
Refs #43

Update content and system error messages using the notes to provide additional context and recommended actions.

### DIFF
--- a/internal/workflow/preprocessing.go
+++ b/internal/workflow/preprocessing.go
@@ -76,16 +76,13 @@ func (r *PreprocessingWorkflowResult) newEvent(ctx temporalsdk_workflow.Context,
 func (r *PreprocessingWorkflowResult) validationError(
 	ctx temporalsdk_workflow.Context,
 	ev *eventWrapper,
-	msg string,
-	failures []string,
+	msg ...string,
 ) {
 	r.Outcome = OutcomeContentError
 	ev.Complete(
 		ctx,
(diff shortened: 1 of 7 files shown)


djjuhasz:
Here's an screenshot of the validate SIP name validation error message for an example of how the new messages look:
Screenshot from 2025-03-28 09-27-28

Note that I have not added the "Your SIP has been moved to the failed-sips directory" line to any of the error messages — I think it would be better to report this to the user from the activity that moves the failed SIP.

internal/workflow/preprocessing.go
```diff
@@ -662,3 +849,16 @@ func writePREMISFile(ctx temporalsdk_workflow.Context, sip sip.SIP, veraPDFVersi
 
 	return nil
 }
+
+func markdownUList(items []string) string {
```
jraddaoui, line 853: I don't get the name, what's the meaning of the `U`? I also don't see a lot of Markdown, I'd just call it `formatList` or something like that.

jraddaoui (approved):
Really nice, thanks @djjuhasz!

internal/workflow/preprocessing.go
djjuhasz: Markdown Unordered List, like the <ul> HTML element.

djjuhasz: I changed the function name to `ul()` and added a comment describing its functionality.

PR #139, opened 2025-03-28
Don't check file formats for migrated SIPs

### LINKED ISSUE #135: Problem: DigitizedAIPs and BornDigitalAIPs should not have formats disallowed
**Describe the bug**

The migration packages (identified as BornDigitalAIP or DigitizedAIP) include file formats that are not permitted according to SFA's current rules. When one of these SIPs is ingested, it will fail at **Validate SIP file formats**. However, SFA has decided that they will address disallowed file formats in the migration packages as a separate, post-migration project.

**To Reproduce**

1. Ingest a migration package that contains a disallowed format (e.g. a quicktime file)
2. See it fail!

**Expected behavior**

If a transfer is identified as a BornDigitalAIP or DigitizedAIP, all of the files included in the transfer should be permitted to pass through into Archivematica. **Validate SIP file formats** should not run for BornDigitalAIPs or DigitizedAIPs.

**Screenshots**

**Additional context**

djjuhasz:
Fixes #135.

Don't check allowed file formats for DigitizedAIP and BornDigitalAIP SIP types.

### DIFF
--- a/internal/workflow/preprocessing.go
+++ b/internal/workflow/preprocessing.go
@@ -326,27 +326,30 @@ func (w *PreprocessingWorkflow) Execute(
 		checksumEv.Succeed(ctx, "SIP checksums match file contents")
 	}
 
-	// Verify that SIP file formats are on allowlist.
-	ev = result.newEvent(ctx, "Validate SIP file formats")
-	var ffvalidateResult ffvalidate.Result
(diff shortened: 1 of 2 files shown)


jraddaoui (approved):
:ok_hand:

Issue #152, opened 2025-06-02
Problem: PREMIS agent for VeraPDF created even when VeraPDF doesn't run


sallain:
**Describe the bug**

A PREMIS agent is being created for VeraPDF even when VeraPDF doesn't run. The resulting premis.xml file contains the following agent:

`<premis:agent><premis:agentIdentifier><premis:agentIdentifierType valueURI="http://id.loc.gov/vocabulary/identifiers/local">url</premis:agentIdentifierType><premis:agentIdentifierValue>https://verapdf.org</premis:agentIdentifierValue></premis:agentIdentifier><premis:agentName>veraPDF 1.26.2</premis:agentName><premis:agentType>software</premis:agentType></premis:agent></premis:premis>`

In Archivematica, when AM tries to load the premis.xml file, it encounters the agent but can't find any associated events, so it throws an error.

**To Reproduce**

1. Upload a SIP that does not contain PDF/A files (e.g. a sample containing only jp2 files)
2. Notice that the Archivematica task _Load PREMIS events from metadata/premis.xml_ has failed.
3. In Archivematica, you can click into the task for more information - it says "The premis:agent element with premis:agentIdentifierType='url' and premis:agentIdentifierValue='https://verapdf.org' is not related to any premis:event element"
4. In the premis.xml file (you can find it in the resulting AIP), look at the last line and see the PREMIS agent value as written above.

**Expected behavior**

If there are no PDF/As in the package, then _Validate SIP files_ should not run. Therefore, no PREMIS agent should be created.

**Additional context**

Related to #63, #118

fiver-watson:
Uploaded a SIP with no PDFs, did not find verapdf anywhere in the METS! And also no longer seeing the error writing the Enduro PREMIS file to the AM METS. Noice!

PR #154, opened 2025-06-05
Add file validator scope

djjuhasz:
This was prompted by work on #152 but is mostly unrelated.

This change recognizes that veraPDFValidator operates on a directory, rather than an individual file, while allowing for future validators that operate on a single file.

The added `fvalidate.Scope()` method returns the intended scope of the `Validate()` method, either a single file or a directory of files.

The added 'veraPDFValidator.scope' property indicates that the validator operates on a directory of files.

### DIFF
--- a/internal/activities/validate_files.go
+++ b/internal/activities/validate_files.go
@@ -6,10 +6,11 @@ import (
 	"fmt"
 	"slices"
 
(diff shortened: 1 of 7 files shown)


jraddaoui (approved):
:ok_hand:

Issue #164, opened 2025-10-06
Feature: ensure that schema version is 4.0 or 5.0


sallain:
**Is your feature request related to a problem? Please describe.**

Only eCH-0160 4.0 and 5.0 are acceptable. The version of the schema is encoded in `<schemaVersion>` in the metadata.xml file.

**Describe the solution you'd like**

Add a check on the schema version. The value should be either `"4.0"` or `"5.0"` - not `"5.1"`, etc.

**Describe alternatives you've considered**

None

**Additional context**

None

PR #166, opened 2025-12-01
Parse and validate manifest schema version

jraddaoui:
Refs #164.


internal/manifest/manifest_test.go
```diff
@@ -128,107 +128,113 @@ var (
...
 			name:   "Returns an empty list from an empty manifest",
 			reader: strings.NewReader(""),
-			want:   manifest.Manifest{},
+			want:   &manifest.Manifest{Files: map[string]*manifest.File{}},
 		},
 		{
 			name:    "Errors on a missing closing tag",
 			reader:  strings.NewReader(`<datei id="_ZSANrSklQ9HGn99yjlUumz"><name>arelda.xsd</name>`),
-			want:    manifest.Manifest{},
+			want:    nil,
```
djjuhasz: You can just remove the `want: nil` entirely, it should be equivalent.

PR #188, opened 2026-04-08
Check for the correct extraction directory name

### LINKED ISSUE #169: Problem: no error if SIP is missing top-level directory
**Describe the bug**

All SIPs should be structured like this, generally speaking:

```
my-package.zip
└── my-package
    ├── content
    └── header
```

Note the top-level directory, `my-package`.

SFA have reported that if the top-level directory is missing, the SIP then fails at the name check with the following error

Image

While this error is correct (the name is NOT compliant!), the SIP should fail sometime earlier, with a more specific error message.

**To Reproduce**

1. Make a zipped package where the top-level directories are `content` and `header`
2. Run it

**Expected behavior**

Fail earlier in the workflow, possibly at `Identify SIP structure`?

**Screenshots**

**Additional context**

djjuhasz:
Fixes #169.

- Add the fields BatchID, SIPID, and SIPName to the workflow params passed from the parent ingest workflow
- After extracting the SIP archive file, check that the extract directory contains a top-level directory with the same name as the original ZIP file name.

### DIFF
--- a/internal/workflow/preprocessing.go
+++ b/internal/workflow/preprocessing.go
@@ -12,6 +12,8 @@ import (
 	"github.com/artefactual-sdps/temporal-activities/bagvalidate"
 	"github.com/artefactual-sdps/temporal-activities/ffvalidate"
 	"github.com/artefactual-sdps/temporal-activities/xmlvalidate"
+	"github.com/google/uuid"
+	"go.artefactual.dev/tools/fsutil"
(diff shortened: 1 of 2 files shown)


jraddaoui (approved):
LGTM 👌!

djjuhasz:
I've updated the PR to continue validation if the extraction directory check fails.

PR #190, opened 2026-04-10
Reduce APIS pollInterval values in dev

djjuhasz:
Reduce APIS pollInterval values from 30s to 1s in the kube overlays to reduce preprocessing workflow run times in development. Because the APIS mock server requires 3 polls to complete analysis this reduces the API poll activity time from ~1m30s to ~3s.

### DIFF
--- a/hack/kube/overlays/dev/preprocessing-secret.yaml
+++ b/hack/kube/overlays/dev/preprocessing-secret.yaml
@@ -32,7 +32,7 @@ stringData:
     enabled = true
     url = "http://apis-mock.enduro-sdps:8080"
     timeout = "10s"
-    pollInterval = "30s"
+    pollInterval = "1s"
     token = "mock-token"
 
     [apis.oidc]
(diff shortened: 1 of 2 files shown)


jraddaoui (approved):
Thanks @djjuhasz! May be tight, even if it's a mocked server, I'd consider setting it to 5 seconds.

djjuhasz:
@jraddaoui 1 second worked fine when I tested it.  Where do you think that timing might be tight?  The mock server is certainly capable of returning a response much faster than 1 second, and I don't think we'll be making a lot of simultaneous requests.

jraddaoui:
Feels safer to me, and it doesn't add a big amount of time as before. We may add extra processing needs to the mock server, add k8s proxies like linkerd, or other changes that could increase response time. Up to you, we can always change it again ;)

djjuhasz:
Well, I'm impatient so I'll take the risk in exchange for the preprocessing workflow running 12 seconds faster when I'm doing testing. :smiling_imp:  I think the worst thing that can happen on a slow response from the mock server is that the activity will poll an extra time or two, I'd only really be concerned if we were piling a lot of requests too fast and DOS'd the mock server.

Issue #173, opened 2026-02-25
Feature: Add APIS integration to SFA preprocessing workflow


fiver-watson:
NOTE: this issue ticket description was updated 2026-02-26 to reflect updates from SFA

-----

## Background and context

SFA's long-term solution for digital preservation involves many disparate systems across multiple different departments and agencies coordinating. However, the Archival Information System (AIS) is intended to be source of truth for all archival metadata. Given that various users will be creating and submitting metadata that at times may be in conflict with existing records or other submissions, SFA has developed an intermediary application, called **APIS**, to negotiate the review and synchronization of metadata prior to import into the AIS. 

To support this new integration, ASI will need to modify the existing SFA preprocessing child workflow so that it can interact with APIS via API and confirm there are no outstanding conflicts before proceeding with preservation processing. 

## Proposed implementation

After meetings with SFA and third-party developers, we propose the following integration plan for APIS: 

![Image](https://github.com/user-attachments/assets/af604da8-ba5e-4130-bc11-b854c052eb7a)

First, preprocessing continues normally until validation has been successful. Any content failures occurring during validation will mean that APIS is never contacted, and the workflow will end as it currently does. 

If validation succeeds, the Enduro will: 

* Make a copy of the SIP's eCH-0160 metadata.xml file
* Deliver that copy to APIS via API, using the `importTasks` POST endpoint
* This endpoint has 3 required parameters:
   * `file`: the eCH-0160 metadata.xml file from the SIP
   * `sipType`: The type of SIP identified earlier in the SFA preprocessing workflow the API's enum has been aligned to use the same sip types as we do in Enduro:
      * DigitizedAIP
      * DigitizedSIP
      * BornDigitalAIP
      * BornDigitalSIP   
   * `username`: the user logged in who is currently submitting the request. String; can be an email for example. 
* APIS should return a request ID as part of the response. This ID needs to be kept in memory, both for status updates on the current process and to initiate the `importRun` API call later in the related post-storage workflow. 
* As it waits for a final outcome from the request, Enduro can query the status of the previous `importTasks` request using the returned ID
   * A GET `/api/ImportTasks/{id}/status` request using the returned ID can return the following elements: 

```
       "status": "Neu",
       "analysisResult": "AlleGleich",
       "analysisProgressInPercent": 0,
       "importResult": "Erfolgreich",
       "importProgressInPercent": 0,
       "processedDocumentCount": 0,
       "totalDocumentCount": 0
```
[...]

PR #194, opened 2026-04-13
Update APIS spec, mock server, client and activities

jraddaoui:
Add latest apis/openapi3.json in two steps, the first commit adds the spec reorganized to better match the existing structure and the second one updates it as it was provided. Regenerate server and client stubs from the new spec and update mock server implementation and the workflow activities using the generated client.

Refs #173.

### DIFF
--- a/hack/apis-mock/internal/gen/generate.go
+++ b/hack/apis-mock/internal/gen/generate.go
@@ -1,3 +1,3 @@
 package gen
 
-//go:generate go run github.com/ogen-go/ogen/cmd/ogen@v1.20.0 --config ogen.yml --target . --package gen --clean ../../../apis/openapi3.json
+//go:generate go run github.com/ogen-go/ogen/cmd/ogen@v1.20.0 --config ogen.yml --target . --package gen --clean ../../../../apis/openapi3.json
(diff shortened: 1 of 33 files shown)


hack/apis-mock/internal/gen/generate.go
```diff
@@ -1,3 +1,3 @@
 package gen
 
-//go:generate go run github.com/ogen-go/ogen/cmd/ogen@v1.20.0 --config ogen.yml --target . --package gen --clean ../../../apis/openapi3.json
+//go:generate go run github.com/ogen-go/ogen/cmd/ogen@v1.20.0 --config ogen.yml --target . --package gen --clean ../../../../apis/openapi3.json
```
djjuhasz, line 3: Huh, why the change of path?

djjuhasz (approved):
LGTM @jraddaoui. 👍 I just had one little question.

hack/apis-mock/internal/gen/generate.go
jraddaoui: The first time I generated the code it was in a different path, then I moved it, but I forgot to change it here.

PR #195, opened 2026-04-21
Remove standalone Tilt environment

jraddaoui:
Drop the standalone Tilt and Kubernetes development environment and keep only the configuration needed to run this project through Enduro's dev env. Flatten the Kubernetes manifests into a single kustomize root, update the README to describe both child workflows, and replace the Tilt CI workflow with manifest and Tiltfile validation.


jraddaoui:
I'll do the same in other child workflow repositories after we get this one right. It will also require to make a small change in Enduro to load the default `Tiltfile` instead of `Tiltfile.enduro`.

djjuhasz (approved):
Awesome, thanks @jraddaoui. :)

PR #196, opened 2026-04-24
Add APIS decision signaling to preprocessing

jraddaoui:
Signal the parent workflow when APIS reports metadata conflicts and wait for a decision before continuing preprocessing. Keep the APIS task ID and decision as local state, excluding them from the Temporal result.

Refs #173. See https://github.com/artefactual-sdps/enduro/issues/1536 and https://github.com/artefactual-sdps/enduro/pull/1608.

### DIFF
--- a/internal/workflow/preprocessing.go
+++ b/internal/workflow/preprocessing.go
@@ -60,10 +60,18 @@ type (
 		SIPName string
 	}
 
+	// PreprocessingWorkflowResult is returned to Enduro processing workflow and
+	// also serves as this workflow's local state.
 	PreprocessingWorkflowResult struct {
 		Outcome           Outcome
 		RelativePath      string
 		PreservationTasks []*eventlog.Event
+
+		// These APIS fields are kept as local state for now. They are excluded
+		// from Temporal serialization until we decide how they should be shared
+		// with the AIS poststorage workflow.
+		APISTaskID   string `json:"-"`
+		APISDecision string `json:"-"`
 	}
 )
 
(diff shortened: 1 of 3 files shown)


internal/workflow/preprocessing.go
```diff
@@ -767,43 +777,130 @@ func (w *PreprocessingWorkflow) createAPISImportTask(
...
+	if info.ParentWorkflowExecution == nil {
+		result.systemError(
+			ctx,
+			fmt.Errorf(
+				"missing parent workflow execution while waiting for a decision on APIS import task %q",
+				result.APISTaskID,
+			),
+			ev,
+			"submission to APIS has failed.",
+			"Could not coordinate the required human review because the preprocessing workflow has no parent workflow context.",
```
djjuhasz: I think this error message is probably too technical for an operator, and it's missing a suggestion for next steps. Maybe something like this would be better:

"A system error occurred requesting a human decision on how to continue processing. Please try again, or ask a system administrator to investigate."

internal/workflow/preprocessing.go
```diff
@@ -767,43 +777,130 @@ func (w *PreprocessingWorkflow) createAPISImportTask(
...
+	var decision DecisionResponse
+	temporalsdk_workflow.GetSignalChannel(ctx, DecisionResponseSignal).Receive(ctx, &decision)
+	result.APISDecision = decision.Option
+
+	switch result.APISDecision {
+	case DecisionOptionContinueOverwrite, DecisionOptionContinueAppend:
+		ev.Succeed(
+			ctx,
+			fmt.Sprintf(
+				"APIS analysis completed for import task ID %q with result %q after user decision %q",
```
djjuhasz: This message sounds to me like the analysis result "Conflict" was a result of the user decision. I think a message more like the cancel message is clearer, e.g. "APIS detected metadata conflicts for import task ID %q but ingest was continued with user decision %q"

djjuhasz (approved):
LGTM @jraddaoui, I just had a couple of suggested text changes for the UI messages.

jraddaoui:
Thanks @djjuhasz!

Issue #174, opened 2026-02-25
Enhancement: Add APIS integration to SFA post-storage workflow


fiver-watson:
## Background and context

SFA's long-term solution for digital preservation involves many disparate systems across multiple different departments and agencies coordinating. However, the Archival Information System (AIS) is intended to be source of truth for all archival metadata. Given that various users will be creating and submitting metadata that at times may be in conflict with existing records or other submissions, SFA has developed an intermediary application, called APIS, to negotiate the review and synchronization of metadata prior to import into the AIS.

During preprocessing, Enduro and APIS will resolve any potential metadata conflicts and deliver the initial SIP metadata.xml file. In return, APIS will supply an ID for the request. This ID can then be used in post-storage to deliver a copy of the AIP METS.xml file for the AIS to index and store. Details of this earlier preprocessing integration can be found here: 

* #173 

To support the second phase of this new integration, ASI will need to modify the existing SFA post-storage child workflow so that it can use the ID delivered during preprocessing and send the METS file to the AIS via APIS. 

## Implementation proposal

The following workflow diagram describes the proposed initial post-storage workflow modifications at a high level: 

![Image](https://github.com/user-attachments/assets/5f3eb04e-a074-4a0f-bb3d-8f1acb6ede1a)

As usual, during preservation Enduro will continue to poll the preservation engine for updates, until it can confirm that preservation has completed successfully and the PIP has been transformed into an AIP that is now stored.  Once this is confirmed, the post-storage workflow can begin. Enduro will need to: 

* Fetch a copy of the AIP METS.xml file 
* Using the ID value returned from APIS during preprocessing, initiate a POST to `/api/ImportTasks/{id}/importRuns`
* This API request expects 3 required parameters to be included:
   *  `id`: the ID value returned from APIS during preprocessing, via the `importTasks` request
   * `file` the METS.xml file from the resulting AIP
   * `importBehavior`: this pram expects one of two enum values: 
      * `AppendOnly`: This is the default value if the file param is not included in the request (it's the only param here that is optional), and will be used in most cases when no metadata conflict is identified eariler during preprocessing via APIS
[...]

PR #201, opened 2026-05-08
Carry APIS metadata to AIS poststorage

jraddaoui:
Record the APIS import task ID and human review decision in preprocessing custom metadata so the AIS workflow can receive the context it needs.  Read and log that metadata in the AIS poststorage workflow without starting APIS poststorage integration yet.

Refs #174.

### DIFF
--- a/internal/apis/custom_metadata.go
+++ b/internal/apis/custom_metadata.go
@@ -1,0 +1,40 @@
+package apis
+
+import (
+	"encoding/json"
+	"fmt"
+)
+
+const CustomMetadataKey = "apis"
+
+type CustomMetadata struct {
+	ImportTaskID string `json:"importTaskId"`
+	Decision     string `json:"decision"`
+}
+
+func (m CustomMetadata) Marshal() ([]byte, error) {
+	if m.ImportTaskID == "" {
+		return nil, fmt.Errorf("APIS custom metadata requires import task ID")
+	}
+
+	data, err := json.Marshal(m)
+	if err != nil {
+		return nil, fmt.Errorf("marshal APIS custom metadata: %w", err)
+	}
+
+	return data, nil
+}
+
+func (m *CustomMetadata) Unmarshal(data []byte) error {
+	if m == nil {
+		return fmt.Errorf("APIS custom metadata destination is nil")
+	}
+	if err := json.Unmarshal(data, m); err != nil {
+		return fmt.Errorf("unmarshal APIS custom metadata: %w", err)
+	}
+	if m.ImportTaskID == "" {
+		return fmt.Errorf("APIS custom metadata requires import task ID")
+	}
+
+	return nil
+}
--- a/internal/workflow/preprocessing.go
+++ b/internal/workflow/preprocessing.go
@@ -521,7 +521,6 @@ func (w *PreprocessingWorkflow) Execute(
 
 	// Create APIS import task.
 	if w.apisEnabled {
-		// TODO: Add APIS import task ID and decision to result.CustomMetadata.
 		if ok := w.createAPISImportTask(ctx, result, sip); !ok {
 			return result, nil
 		}
@@ -630,10 +629,10 @@ func withFilesysActOpts(ctx temporalsdk_workflow.Context) temporalsdk_workflow.C
 	})
 }
 
-// createAPISImportTask submits metadata to APIS, stores the resulting task ID
-// and decision, and handles the analysis outcome for the current workflow execution.
-// It returns false when processing should stop after recording the failure in
-// the workflow result/state.
+// createAPISImportTask submits metadata to APIS, waits for analysis, and records
+// the resulting custom metadata for the AIS poststorage workflow. It returns
+// false when processing should stop after recording the failure in the workflow
+// result.
 func (w *PreprocessingWorkflow) createAPISImportTask(
 	ctx temporalsdk_workflow.Context,
 	result *childwf.PreprocessingResult,
@@ -668,10 +667,11 @@ func (w *PreprocessingWorkflow) createAPISImportTask(
 		)
 		return false
 	}
+	metadata := apis.CustomMetadata{ImportTaskID: createAPISImportTask.TaskID}
 	task.Succeed(
 		temporalsdk_workflow.Now(ctx),
 		"Submitted metadata to APIS with import task ID %q",
-		createAPISImportTask.TaskID,
+		metadata.ImportTaskID,
 	)
 
 	task = result.NewTask(temporalsdk_workflow.Now(ctx), "Wait for APIS analysis")
@@ -687,7 +687,7 @@ func (w *PreprocessingWorkflow) createAPISImportTask(
 			},
 		}),
 		apis.PollImportTaskStatusActivityName,
-		&apis.PollImportTaskStatusParams{TaskID: createAPISImportTask.TaskID},
+		&apis.PollImportTaskStatusParams{TaskID: metadata.ImportTaskID},
 	).Get(ctx, &pollAPISImportTaskStatus)
 	if err != nil {
 		logger.Error("System error", "message", err.Error())
@@ -705,17 +705,20 @@ func (w *PreprocessingWorkflow) createAPISImportTask(
 		task.Succeed(
 			temporalsdk_workflow.Now(ctx),
 			"APIS analysis completed for import task ID %q with result %q",
-			createAPISImportTask.TaskID,
+			metadata.ImportTaskID,
 			pollAPISImportTaskStatus.AnalysisResult,
 		)
-		return true
 	case apisgen.AnalysisResultKonflikte:
-		return w.waitForAPISDecision(ctx, result, task, createAPISImportTask.TaskID)
+		decision, ok := w.waitForAPISDecision(ctx, result, task, metadata.ImportTaskID)
+		metadata.Decision = decision
+		if !ok {
+			return false
+		}
 	case apisgen.AnalysisResultFehler:
 		logger.Error(
 			"System error",
 			"message",
-			fmt.Sprintf("APIS analysis failed for task %q", createAPISImportTask.TaskID),
+			fmt.Sprintf("APIS analysis failed for task %q", metadata.ImportTaskID),
 		)
 		result.SystemError(
 			temporalsdk_workflow.Now(ctx),
@@ -728,7 +731,7 @@ func (w *PreprocessingWorkflow) createAPISImportTask(
 		logger.Error("System error", "message", fmt.Errorf(
 			"unexpected APIS analysis result %q for task %q",
 			pollAPISImportTaskStatus.AnalysisResult,
-			createAPISImportTask.TaskID,
+			metadata.ImportTaskID,
 		))
 		result.SystemError(
 			temporalsdk_workflow.Now(ctx),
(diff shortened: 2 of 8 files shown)


internal/apis/custom_metadata.go
```diff
@@ -0,0 +1,40 @@
...
+)
+
+const CustomMetadataKey = "apis"
+
+type CustomMetadata struct {
+	ImportTaskID string `json:"importTaskId"`
+	Decision     string `json:"decision"`
+}
+
+func MarshalCustomMetadata(m CustomMetadata) (json.RawMessage, error) {
```
djjuhasz: I think it would be better to implement the `json.Marshaler` interface for marshaling `CustomMetadata`.

internal/apis/custom_metadata.go
```diff
@@ -0,0 +1,40 @@
...
+
+	data, err := json.Marshal(m)
+	if err != nil {
+		return nil, fmt.Errorf("marshal APIS custom metadata: %w", err)
+	}
+
+	return data, nil
+}
+
+func UnmarshalCustomMetadata(data json.RawMessage, dst *CustomMetadata) error {
```
djjuhasz: And implement `json.Unmarshaler` here.

internal/workflow/preprocessing.go
```diff
@@ -638,6 +651,7 @@ func (w *PreprocessingWorkflow) createAPISImportTask(
 	ctx temporalsdk_workflow.Context,
 	result *childwf.PreprocessingResult,
 	sip sip.SIP,
+	metadata *apis.CustomMetadata,
```
djjuhasz: Isn't the custom metadata already included in `childwf.PreprocessingResult`? Why is it necessary to pass `apis.CustomMetadata` as  well?

internal/workflow/preprocessing.go
```diff
@@ -705,17 +720,17 @@ func (w *PreprocessingWorkflow) createAPISImportTask(
...
 			temporalsdk_workflow.Now(ctx),
 			"APIS analysis completed for import task ID %q with result %q",
-			createAPISImportTask.TaskID,
+			metadata.ImportTaskID,
 			pollAPISImportTaskStatus.AnalysisResult,
 		)
 		return true
 	case apisgen.AnalysisResultKonflikte:
-		return w.waitForAPISDecision(ctx, result, task, createAPISImportTask.TaskID)
+		return w.waitForAPISDecision(ctx, result, task, metadata)
```
djjuhasz: Again, shouldn't the metadata be included in the `PreprocessingResult`?

internal/workflow/preprocessing.go
```diff
@@ -748,15 +763,14 @@ func (w *PreprocessingWorkflow) waitForAPISDecision(
 	ctx temporalsdk_workflow.Context,
 	result *childwf.PreprocessingResult,
 	task *childwf.Task,
-	taskID string,
+	metadata *apis.CustomMetadata,
```
djjuhasz: The metadata should be in the result again.

djjuhasz (changes requested):
@jraddaoui I have some questions about how the custom metadata is passed around and marshalled/unmarshalled.

internal/apis/custom_metadata.go
jraddaoui: I considered implementing the `json.Marshaler` and `json.Unmarshaler` interfaces. The implementation would look roughly like this:

```go
func (m CustomMetadata) MarshalJSON() ([]byte, error) {
	if m.ImportTaskID == "" {
		return nil, fmt.Errorf("APIS custom metadata requires import task ID")
	}

	// Use a local alias so json.Marshal uses default struct encoding
	// instead of recursively calling CustomMetadata.MarshalJSON.
	type customMetadata CustomMetadata
	return json.Marshal(customMetadata(m))
}

func (m *CustomMetadata) UnmarshalJSON(data []byte) error {
	if m == nil {
		return fmt.Errorf("APIS custom metadata destination is nil")
	}

	// Use a local alias so json.Unmarshal uses default struct decoding
	// instead of recursively calling CustomMetadata.UnmarshalJSON.
	type customMetadata CustomMetadata
	var decoded customMetadata
	if err := json.Unmarshal(data, &decoded); err != nil {
		return err
	}

	*m = CustomMetadata(decoded)
	if m.ImportTaskID == "" {
		return fmt.Errorf("APIS custom metadata requires import task ID")
	}

	return nil
}
```

My hesitation is that our boundary is an opaque Temporal custom metadata map, `map[string]json.RawMessage`, where we marshal the APIS metadata before passing it into the Temporal payload. At that point, Temporal is not discovering how to marshal `CustomMetadata`; it is just carrying an already marshaled `json.RawMessage`.

For that shape, explicit `MarshalCustomMetadata` and `UnmarshalCustomMetadata` helpers make the boundary and validation point clearer. Implementing the interfaces would mostly help if we expect callers to use generic JSON paths directly.

So I’d lean toward keeping the helpers unless we expect `CustomMetadata` to be marshaled or unmarshaled through generic JSON paths elsewhere.

P.S. I’m not a big fan of the alias type to avoid recursive calls.

internal/workflow/preprocessing.go
jraddaoui: `result.CustomMetadata` is the final opaque payload returned by the workflow: `map[string]json.RawMessage`.

`apis.CustomMetadata` is the typed state we build before writing that final payload. The APIS flow first gets the import task ID, may later add the human decision, and only after that do we serialize it into `result.CustomMetadata`.

internal/apis/custom_metadata.go
djjuhasz: Oh, right.  The recursive nature of `MarshalJSON()` on structs always throws me for a loop.

I still think that following the pattern of `json.Marshal()` and `json.Unmarshal()` is more ergonomic and intuitive, e.g.
```go
func (m CustomMetadata) Marshal() ([]byte, error) {
	if m.ImportTaskID == "" {
		return nil, fmt.Errorf("APIS custom metadata requires import task ID")
	}

	data, err := json.Marshal(m)
	if err != nil {
		return nil, fmt.Errorf("marshal APIS custom metadata: %w", err)
	}

	return data, nil
}

func (m *CustomMetadata) Unmarshal(data []byte) error {
	if m == nil {
		return fmt.Errorf("APIS custom metadata destination is nil")
	}
	if err := json.Unmarshal(data, m); err != nil {
		return fmt.Errorf("unmarshal APIS custom metadata: %w", err)
	}
	if m.ImportTaskID == "" {
		return fmt.Errorf("APIS custom metadata requires import task ID")
	}

	return nil
}
```

Then
```go
data, err := apis.MarshalCustomMetadata(apis.CustomMetadata{
	ImportTaskID: "task-000001",
	Decision:     "Continue and append",
})
```

Becomes: 
```go
data, err := apis.CustomMetadata{
	ImportTaskID: "task-000001",
	Decision:     "Continue and append",
}.Marshal()
```

And

```go
err := apis.UnmarshalCustomMetadata(
	json.RawMessage(`{"importTaskId":"task-000001","decision":"Continue and append"}`),
	&metadata,
)
```

becomes:

```go
err := metadata.Unmarshal(
	json.RawMessage(`{"importTaskId":"task-000001","decision":"Continue and append"}`),
)
```

djjuhasz: By the way, I tested `CustomMetadata.Marshal()` and `CustomMetadata.Unmarshal()` in the Enduro code, using the `[]byte` type (as written above) instead of `json.RawMessage`  and there were no type casting errors.

internal/workflow/preprocessing.go
djjuhasz: Okay, but I don't see a reason to pass a pointer to `metadata` instead of just returning the `ImportTaskID`. I think a more natural and self-descriptive interface would be something like:

```
type APISImportResult struct {
	ImportTaskID string
	Continue     bool
}

// createAPISImportTask submits metadata to APIS, stores the resulting task ID
// and decision, and handles the analysis outcome for the current workflow execution.
// It returns false when processing should stop after recording the failure in
// the workflow result/state.
func (w *PreprocessingWorkflow) createAPISImportTask(
	ctx temporalsdk_workflow.Context,
	result *childwf.PreprocessingResult,
	sip sip.SIP,
) APISImportResult {
...
```

You could also directly return `string, bool` (ImportTaskID, continue) but that's not as self-descriptive as using a result struct.

Or you could just return `string, err` and decide whether to continue in the caller based on the returned error value.

internal/workflow/preprocessing.go
```diff
@@ -521,10 +521,23 @@ func (w *PreprocessingWorkflow) Execute(
 
 	// Create APIS import task.
 	if w.apisEnabled {
-		// TODO: Add APIS import task ID and decision to result.CustomMetadata.
-		if ok := w.createAPISImportTask(ctx, result, sip); !ok {
+		var metadata apis.CustomMetadata
+		if ok := w.createAPISImportTask(ctx, result, sip, &metadata); !ok {
 			return result, nil
 		}
+		data, err := apis.MarshalCustomMetadata(metadata)
```
djjuhasz: Okay, you claim you need to pass `&metadata` so you can get the unencoded `apis.CustomMetadata` value but the only thing you are doing with `metadata` is to immediately encode it and then later assign it to `result.CustomMetadata` — so why is it better to do that here, rather than in `createAPISImportTask()`? 😆

internal/workflow/preprocessing.go
jraddaoui: We need the two fields, the task ID and the decision, that's why I was using the `apis.CustomMetadata` struct. I'll return the struct instead of passing it as a pointer. The boolean instead of an error is nice because we never return errors, we collect them in the result.

internal/apis/custom_metadata.go
jraddaoui: Sure, I'll do it. `json.RawMessage` is just `[]byte`.

internal/workflow/preprocessing.go
djjuhasz: `metadata.decision` is not set anywhere in `createAPISImportTask()` so I don't understand why you need it for this function call.

djjuhasz: You also aren't setting `metadata.decision` in the caller of `createAPISImportTask()`.

internal/apis/custom_metadata.go
jraddaoui: See 4e6a67c.

internal/workflow/preprocessing.go
jraddaoui: See b1b6b63.

internal/workflow/preprocessing.go
jraddaoui: I see, let me do that too :wink:

internal/workflow/preprocessing.go
djjuhasz: Oops, I missed that you were calling `waitForAPISDecision()` in `createAPISImportTask()`.  I see how passing `&metadata` to `waitForAPISDecision()` makes sense, but I still don't understand why it needs to be returned from `createAPISImportTask()`.

jraddaoui: Done in 719811a.

internal/workflow/preprocessing.go
jraddaoui: Done in 719811a.

jraddaoui:
Thanks @djjuhasz! I'll squash before merging.

internal/workflow/preprocessing.go
```diff
@@ -668,10 +668,11 @@ func (w *PreprocessingWorkflow) createAPISImportTask(
 		)
 		return false
 	}
+	metadata.ImportTaskID = createAPISImportTask.TaskID
```
djjuhasz: We can define `metadata` here and skip the earlier declaration: e.g. 

```go
metadata := apis.CustomMetadata{ImportTaskID: createAPISImportTask.TaskID}
```

djjuhasz (approved):
Looks great @jraddaoui, thanks.  I added one comment about where `metadata` is defined, but it's not a big deal.

PR #203, opened 2026-05-11
Remove no longer needed steps from poststorage

jraddaoui:
Refs #174.


jraddaoui:
@djjuhasz, ignore the changes related to the get AIP path activity. I realized later that we don't need it to fetch the METS file, so I'm removing it in #205.

djjuhasz (approved):
LGTM 👍

PR #204, opened 2026-05-11
Use ssclient for Storage Service activities

jraddaoui:
Replace the custom AMSS client with go.artefactual.dev/ssclient and wire poststorage activities to the package service directly. Update poststorage AMSS config to use ssclient.Config fields directly.

Parse the AIP UUID once at workflow startup before scheduling any activities, then pass the parsed UUID through AMSS activity parameters.

Based on #203. Refs #174.


jraddaoui:
@djjuhasz, ignore the changes related to the get AIP path activity. I realized later that we don't need it to fetch the METS file, so I'm removing it in #205.

djjuhasz (approved):
LGTM 👍

PR #205, opened 2026-05-11
Remove AIP path lookup from poststorage

jraddaoui:
Drop the GetAIPPath activity and its worker registration because poststorage no longer needs Storage Service package metadata before fetching the METS file. Build the temporary directory and METS relative path from the parsed AIP UUID instead.

Based on #204. Refs #174.


djjuhasz (approved):
LGTM 👍

PR #206, opened 2026-05-12
Add APIS import run delivery to poststorage

jraddaoui:
Create poststorage tasks for downloading the AIP METS, submitting it to
APIS as an import run, and waiting for the APIS import result. Require
APIS custom metadata when APIS is enabled, map preprocessing decisions
to APIS import run behaviour.

Refs #174.

### DIFF
--- a/internal/workflows/poststorage.go
+++ b/internal/workflows/poststorage.go
@@ -15,21 +15,27 @@ import (
 
 	"github.com/artefactual-sdps/preprocessing-sfa/internal/amss"
 	"github.com/artefactual-sdps/preprocessing-sfa/internal/apis"
+	apisgen "github.com/artefactual-sdps/preprocessing-sfa/internal/apis/gen"
 	"github.com/artefactual-sdps/preprocessing-sfa/internal/config"
 )
 
 type Poststorage struct {
-	cfg config.PoststorageConfig
+	cfg         config.PoststorageConfig
+	apisEnabled bool
 }
 
-func NewPoststorage(cfg config.PoststorageConfig) *Poststorage {
-	return &Poststorage{cfg: cfg}
+func NewPoststorage(cfg config.PoststorageConfig, apisEnabled bool) *Poststorage {
+	return &Poststorage{
+		cfg:         cfg,
+		apisEnabled: apisEnabled,
+	}
 }
 
 func (w *Poststorage) Execute(
 	ctx temporalsdk_workflow.Context,
 	params *childwf.PostStorageParams,
 ) (r *childwf.PostStorageResult, e error) {
+	r = &childwf.PostStorageResult{}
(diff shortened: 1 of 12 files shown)


internal/workflows/poststorage.go
```diff
@@ -15,65 +15,49 @@ import (
...
-			},
-		),
-		amss.GetAIPPathActivityName,
-		&amss.GetAIPPathActivityParams{
-			AIPUUID: aipUUID,
-		},
-	).Get(ctx, &getAIPPathResult)
-	if err != nil {
+	var apisMetadata apis.CustomMetadata
+	if err := apisMetadata.Unmarshal(data); err != nil {
```
djjuhasz: I think it would be nice to have a helper function to get and unmarshal the apisMetadata, to reduce the cognitive load in the main workflow.  Something like `getApisMetadata(params.CustomMetadata) (apis.CustomMetadata, error)`.

internal/workflows/poststorage.go
```diff
@@ -168,8 +188,110 @@ func (w *Poststorage) SessionHandler(ctx temporalsdk_workflow.Context, aipUUID u
...
+		logger.Error("System error", "message", e.Error())
+		result.SystemError(
+			temporalsdk_workflow.Now(ctx),
+			task,
+			"AIP METS download has failed.",
+			"The AIP is stored, but the poststorage workflow failed while downloading the AIP METS file. Please try again, or ask a system administrator to investigate.",
+		)
+		return nil
 	}
+	task.Succeed(temporalsdk_workflow.Now(ctx), "AIP METS downloaded")
```
djjuhasz: Are we cleaning up the downloaded METS file anywhere?

djjuhasz (approved):
LGTM @jraddaoui, but it's a big PR and I had hard time keeping track of everything.  Hopefully I didn't miss any problems. 🤞 

I did have a couple of minor notes, but they aren't blockers.

internal/workflows/poststorage.go
jraddaoui: I'll do it!

internal/workflows/poststorage.go
jraddaoui: Not the easiest to see, but we download into a local dir that is added to the remove paths, and deleted in the defer function: 

- https://github.com/artefactual-sdps/preprocessing-sfa/blob/dev/issue-174-apis-poststorage/internal/workflows/poststorage.go#L175-L178
- https://github.com/artefactual-sdps/preprocessing-sfa/blob/dev/issue-174-apis-poststorage/internal/workflows/poststorage.go#L187
- https://github.com/artefactual-sdps/preprocessing-sfa/blob/dev/issue-174-apis-poststorage/internal/workflows/poststorage.go#L126-L138

jraddaoui:
Thanks @djjuhasz!

PR #207, opened 2026-05-12
Revert "Remove AIP path lookup from poststorage"

jraddaoui:
This reverts commit a7acc7a5fa358dbb68295c3e3dc69febaf8a95ec.

Refs #174.

Sorry for the back and forward @djjuhasz, I thought this was only needed to create the bundle name, but it's actually needed to fetch the METS.


djjuhasz (approved):
👍

PR #208, opened 2026-05-12
Update APIS mock server

jraddaoui:
- Bump Go from 1.25.0 to 1.26.3
- Allow conflict tasks to start import runs
- Configure mock results with env vars

Refs #174.

### DIFF
--- a/hack/apis-mock/README.md
+++ b/hack/apis-mock/README.md
@@ -72,8 +72,8 @@ The mock is intentionally small:
 - one in-memory task record per `POST /api/ImportTasks`
 - one optional import run per task
 - `GET /api/ImportTasks/{id}/status` is the only endpoint that advances state
-- the default path is the happy path, and request hints override it when a test
-  or manual run needs a conflict or failure
+- terminal analysis and import results are configured with environment
+  variables
 
 That shape keeps the implementation easy to read and easy to tweak for local
 development, while still matching the workflow semantics described in the APIS
@@ -118,8 +118,7 @@ go run ./cmd/mockapi
 Server listens on `:8080` by default.
 
 The generated security model expects a bearer token on all endpoints. The
-health endpoint is intentionally not implemented because the current
-integration work only needs the import-task endpoints.
+health endpoint returns a minimal `OK` response.
 
 ## Build binary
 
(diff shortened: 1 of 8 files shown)


hack/apis-mock/cmd/mockapi/main.go
```diff
@@ -1,33 +1,64 @@
...
+		return "", "", fmt.Errorf("%s: %w", envMockImportResult, err)
+	}
+
+	return analysisResult, importResult, nil
+}
+
 func envOrDefault(key, fallback string) string {
-	if v, ok := os.LookupEnv(key); ok && v != "" {
-		return v
+	if v, ok := os.LookupEnv(key); ok && strings.TrimSpace(v) != "" {
```
djjuhasz: I think it's better to use `os.GetEnv(key)` here because we don't need to distinguish whether the environment variable is unset or empty.

hack/apis-mock/internal/mock/server.go
```diff
@@ -33,17 +38,17 @@ func (s *Security) HandleSmart(ctx context.Context, _ gen.OperationName, t gen.S
...
 type Handler struct {
-	mu       sync.Mutex
-	nextTask int
-	nextRun  int
-	tasks    map[string]*taskState
+	mu                    sync.Mutex
+	nextTask              int
+	nextRun               int
+	tasks                 map[string]*taskState
+	defaultAnalysisResult gen.AnalysisResult
```
djjuhasz: Why are these fields named `defaultAnalysisResult` and `defaultImportResult`, and not just `analysisResult` and `importResult`?  The default values `DefaultAnalysisResult` and `DefaultImportResult` are exported, so it seems like they can be used in cases where the defaults are required.

hack/apis-mock/README.md
```diff
@@ -132,24 +131,40 @@ go build -o bin/mockapi ./cmd/mockapi
...
+  -e MOCK_ANALYSIS_RESULT=Konflikte \
+  -e MOCK_IMPORT_RESULT=Fehler \
+  apis-mock:latest
 ```
 
 ## Mock tuning switches
 
 - `MOCK_AUTH_TOKEN=some-token` -> expected bearer token (default: `mock-token`)
 - `PORT=9090` -> change listening port
+- `MOCK_ANALYSIS_RESULT=Konflikte` -> default terminal `analysisResult`
```
djjuhasz: I'm confused by the "default" in "default terminal `analysisResult` here. Can the set `MOCK_ANALYSIS_RESULT` value be overridden somewhere else in the mock server?

hack/apis-mock/README.md
```diff
@@ -132,24 +131,40 @@ go build -o bin/mockapi ./cmd/mockapi
...
+  for import tasks (default: `AlleNeu`; valid values: `AlleNeu`, `AlleGleich`,
+  `Konflikte`, `Fehler`)
+- `MOCK_IMPORT_RESULT=Fehler` -> default terminal `importResult` for import
```
djjuhasz: Ditto on the "default".

djjuhasz (changes requested):
@jraddaoui I have a few questions and notes.

hack/apis-mock/README.md
jraddaoui: No, it can't be overridden. I have updated these comments.

jraddaoui:
Thanks @djjuhasz!

djjuhasz (approved):
Looks great! Thanks @jraddaoui. 👍

PR #210, opened 2026-05-13
Update APIS spec and integration

jraddaoui:
Updates to the APIS API spec and the implementation after testing against a real instance. See individual commits for more details.

- Bump github.com/ogen-go/ogen from 1.20.0 to 1.20.3
- Allow "Analysiert" as import in progress status
- Fix sipType and importBehaviour in APIS spec
- Handle APIS problem JSON import run errors 

Refs #174.


djjuhasz (approved):
LGTM! Thanks @jraddaoui. 👍

PR #214, opened 2026-05-18
Manage ogen with bine

jraddaoui:
Move APIS code generation to Makefile targets that use the ogen binary managed by bine. Remove the go:generate stubs so APIS generation has only one supported entry point through make.


djjuhasz (approved):
👍

Issue #216, opened 2026-05-20
Feature: Add custom workflow API endpoints to support DIP creation workflow


fiver-watson:
Part of the DPS' responsibility will be to deliver packages to Online-Zugang (OZ), the public access portal by which  users can request DIPs. OZ will handle normalization, but the DPS will be responsible for: 

* Receiving a document key (`docKey` - an identifier) for the request from OZ
* Using this docKey to query the AIS and get a related AIP UUID
* Using the AIP UUID to fetch the relevant package from AMSS
* Using the docKey to request the creation and delivery of a metadata.xml file from the AIS
* Parsing the metadata.xml file to determine if the whole AIP is needed, or just certain files
* Creating a package for delivery that follows the eCH structure outlined in the supplied metadata.xml file
* Depositing the package in an object store for OZ to pick up

This means that the DPS will be responsible for negotiating the interactions between all the participating systems (OZ, AIS, AMSS) - but it will not be initiating the workflow. 

## Proposed solution

To allow OZ to initiate the workflow and check for status updates as it progresses, a specification for this workflow is being prepared by a cross-team group, here: 

* [SFA DIP creation and delivery specification](https://docs.google.com/document/d/16w6VxPBcvf6HPqYcWjlzRdckAZfVp5LrhoHQp8EI-ho/edit?tab=t.0)

The high-level workflow is depicted in the following flow diagram, first agreed upon at the 2024 on-site meetings, and reviewed / approved by the current DIP delivery working group:

<img width="7365" height="2172" alt="Image" src="https://github.com/user-attachments/assets/65f5c03a-f438-4811-b45a-b0e7e890672a" />

ASI must design and prepare any necessary API endpoints to support the initial part of this workflow, so that OZ can:

- contact the DPS
- request a package by supplying a docKey 
- receive a workflow ID in response
- use that workflow ID to request status updates as it waits for the package delivery
- receive a deposit location with the requested package created

I will create additional issue tickets as needed for other parts of this proposed workflow. **The deliverable here is the API specification for the DPS** that will allow OZ to initiate the proposed workflow.

fiver-watson:
Updated the issue ticket to replace the now-deprecated 1.0 version of the workflow diagram with the updated v1.1 version. 

Diagram originals are in SFA Miro board [here](https://miro.com/app/board/uXjVMlYiVgs=/?moveToWidget=3458764602096379689&cot=14).

aseles13:
:eyes:

[...]

PR #218, opened 2026-05-27
Add DIPs API design and tooling

jraddaoui:
Define the initial DIPs service contract with create and show methods, bearer authentication, validation, and HTTP mappings. Generate the Goa client, server, CLI, and OpenAPI assets from the design.

Add the Goa and jq tool wiring needed to regenerate the API, and mark generated API directories as linguist-generated.

Design initially created with Codex (gpt-5.5 xhigh) using the [goa-service-designer](https://github.com/goadesign/goa/tree/v3/skills) skill.

Refs #216.


jraddaoui:
@djjuhasz, I gave using RFC 9457 for error responses a try. It was not easy to define a single ProblemDetails schema in the specification that could be used for all error responses, while also having specific examples for each response type, but I think I made it work.

At least on the schema side, I would like to check how the implementation looks before sending anything, but if you have any further feedback on the design or the schema, it would be nice to hear it already.

internal/dips/api/gen/http/openapi.json
djjuhasz, line 12: What is the "application/gob" type?

internal/dips/api/design/design.go
```diff
@@ -45,6 +45,46 @@ var ObjectKey = Type("ObjectKey", String, func() {
...
+	problemTypeInternal     = "/problems/internal-error"
+)
+
+var ProblemDetails = Type("ProblemDetails", func() {
+	Description("RFC 9457 problem details object.")
+	// Use the RFC 9457 "type" member as Goa's error discriminator instead of
+	// adding a private "name" field. This keeps the wire body and OpenAPI schema
+	// to the standard problem details fields while still letting Goa route
+	// shared errors to distinct HTTP status codes.
+	ErrorName("type", String, "URI reference that identifies the problem type.", func() {
```
djjuhasz: It looks like the RFC 9457 standard prefers the type [about:blank](https://datatracker.ietf.org/doc/html/rfc9457#name-aboutblank) when the problem type maps cleanly to an HTTP status response code.  Is it possible to make the type "about:blank" for the four problem types defined without messing up the Goa ErrorName -> generated spec functionality? If necessary, using a private "name" field as a discriminator sounds okay to me. 🤔

internal/dips/api/design/design.go
```diff
@@ -45,6 +45,46 @@ var ObjectKey = Type("ObjectKey", String, func() {
...
+		Enum(problemTypeUnauthorized, problemTypeNotValid, problemTypeNotFound, problemTypeInternal)
+		Example(problemTypeNotValid)
+	})
+	Attribute("title", String, "Short, human-readable summary of the problem type.", func() {
+		Example("Invalid request parameters.")
+	})
+	Attribute("detail", String, "Human-readable explanation specific to this problem occurrence.", func() {
+		Example("The request body is missing the docKey field.")
+	})
+	Attribute("status", Int, "HTTP status code generated by the origin server.", func() {
```
djjuhasz: I think we should use `Int32` here to avoid the generated spec defaulting to `int64`.

internal/dips/api/design/design.go
djjuhasz: I'd say "... and retrieve DIP details" as whether the request is part of a poll is a client-side implementation detail.

djjuhasz: "DIP identifier" isn't a sentence.  We could either go with a full sentence "DIPID uniquely identifies a DIP." or leave it as a sentence fragment and remove the terminating period.

djjuhasz: Also not a sentence

djjuhasz: ditto

djjuhasz: ditto

djjuhasz: ditto

internal/dips/api/design/design.go
```diff
@@ -45,6 +45,46 @@ var ObjectKey = Type("ObjectKey", String, func() {
...
+		// The instance field is optional in RFC 9457, but we include it in the
+		// schema for completeness. For now, we don't generate instance values in
+		// the implementation, so we remove it from the example to avoid confusion.
+		Meta("openapi:example", "false")
+	})
+	Required("type", "title", "status", "detail")
+})
+
 var CreateDIPResult = Type("CreateDIPResult", func() {
 	Description("DIP creation request result.")
```
djjuhasz: This is not a sentence.

internal/dips/api/design/design.go
```diff
@@ -64,17 +104,45 @@ var DIP = Type("DIP", func() {
 })
 
 var _ = Service("DIPs", func() {
-	Description("Requests DIP creation and reports their details.")
+	Description("Requests DIP creation and reports DIP details.")
```
djjuhasz: This is not a sentence.

djjuhasz (changes requested):
This looks great to me @jraddaoui, I just flagged some nitpicky details mostly.  The main thing I would like to see changed is using the "about:blank" problem type as recommended by RFC 9457, but I don't know if this is difficult to do because of the way Goa maps error type definitions to the generated spec.

internal/dips/api/design/design.go
```diff
@@ -64,17 +104,45 @@ var DIP = Type("DIP", func() {
...
 	Security(BearerAuth)
 
-	Error("unauthorized", ErrorResult, "Missing or invalid bearer token.")
-	Error("internal_error", ErrorResult, "Unexpected server error.", func() {
-		Fault()
+	Error(problemTypeNotValid, ProblemDetails, "Invalid request parameters.", func() {
```
djjuhasz: This is not a sentence either, but I'd just remove the terminating period in this case. I don't think the title is meant to be a complete sentence.

internal/dips/api/gen/http/openapi.json
jraddaoui: I believe is related to https://pkg.go.dev/encoding/gob.

djjuhasz: Huh, weird. It's not a registered MIME type: https://www.iana.org/assignments/media-types/media-types.xhtml#application :confused:

internal/dips/api/design/design.go
jraddaoui: That was my initial approach, but the resulting schema did not look right. After reviewing RFC 9457 more closely, I think using our own problem types is the better fit here.

`about:blank` is intended for problems that have no additional semantics beyond the HTTP status code, which is probably enough for our current error cases. But RFC 9457 also allows relative URI references and notes that problem type URIs do not need to resolve immediately; using `/problems/...` gives us stable identifiers now and leaves room to publish documentation for them later, and changing the URI at that point would create a new problem type identity.

When I tried using `about:blank` with a private Goa discriminator, that leaked implementation specific fields into the generated schema and produced multiple almost duplicate refs. Using the RFC type field as the discriminator keeps the response body standard and preserves a single `#/components/schemas/ProblemDetails`.

That said, if you feel strongly about `about:blank`, I can give it another try.

jraddaoui:
Thanks @djjuhasz! Ready for another look.

internal/dips/api/design/design.go
djjuhasz: I went off and tried to figure out an alternate way of implementing RFC 9457 with Goa, and have gotten this far: https://github.com/artefactual-sdps/preprocessing-sfa/tree/dev/issue-216-wip-alt-dips-api-design

I haven't had time to fix up all the examples for the various error types, but I think the potential of implementing extended schemas like https://problems-registry.smartbear.com/missing-request-parameter/ is interesting.

For error responses like "500 Internal Server Error" and "404 Not Found" I don't think there is much supplemental information we can add to the response, so a type of "about:blank" seems appropriate.  Adding more information to identify missing or invalid parameters seems very useful though.

jraddaoui: Looking at https://github.com/artefactual-sdps/preprocessing-sfa/commit/c19f9559a1f610e277158919f8a4a37891c23f21, I think you feel strongly about not using our own type URIs. And that you are also looking at options to extend the problem details in some cases.

I think we should not go that far for now, especially in this small API, where we are only expecting a possible bad request and a single not found. Although I appreciate you looking into that, it may be worth it when we implement this in the Enduro API.

For now, I went back to my initial approach in 232ecf5 and, after some fighting, I have been able to:

- Use `about:black` as the type in all problem details responses. No custom error names or discriminator needed.
- Have a single `ProblemDetails` schema in the spec, while allowing specific examples for each response type.

Not using custom types as discriminators may have some benefits on the implementation side too, allowing the use of `MakeNotValid` and similar helpers, but I'd have to verify that when we get to the implementation part.

internal/dips/api/design/design.go
```diff
@@ -0,0 +1,197 @@
...
+})
+
+var NotFoundProblem = Type("NotFoundProblem", func() {
+	Description("NotFoundProblem represents a missing resource problem.")
+	problemDetailsType("Not Found", "The requested DIP does not exist.", StatusNotFound)
+})
+
+var InternalProblem = Type("InternalProblem", func() {
+	Description("InternalProblem represents an unexpected server problem.")
+	problemDetailsType("Internal Server Error", "The DIP request could not be processed.", StatusInternalServerError)
```
djjuhasz: I think it's better if the help text explicitly states that an error occurred, to be clear that the request itself isn't the reason it can't be processed. E.g. "An error occurred when attempting to process the DIP request."

internal/dips/api/design/design.go
```diff
@@ -0,0 +1,197 @@
...
+	Field(6, "completed_at", DateTime, "The completed_at field contains the time when DIP processing completed.")
+	Field(7, "object_key", ObjectKey, "The object_key field contains the object store key for the completed DIP.")
+	Required("id", "docKey", "status", "created_at")
+})
+
+var _ = Service("DIPs", func() {
+	Description("The DIPs service requests DIP creation and retrieves DIP details.")
+	Security(BearerAuth)
+
+	Error("not_valid", NotValidProblem, "The request parameters are invalid.")
```
djjuhasz: The "not_valid" error name is displayed in the generated schema, which conflicts with the "Bad Request" title and status code. I think it's better to rename this error "bad_request" / "BadRequestProblem" if we aren't extending it beyond the default ProblemDetails fields.

internal/dips/api/gen/http/openapi3.json
```diff
@@ -0,0 +1,365 @@
...
+                }
+              }
+            },
+            "description": "not_valid: The request parameters are invalid."
+          },
+          "401": {
+            "content": {
+              "application/problem+json": {
+                "example": {
+                  "detail": "The Authorization header is missing or invalid.",
```
djjuhasz: The generated description below says "unauthorized: The bearer token is missing or invalid.", I think that description is more accurate?

internal/dips/api/gen/http/openapi3.json
```diff
@@ -0,0 +1,365 @@
...
+                  "$ref": "#/components/schemas/ProblemDetails"
+                }
+              }
+            },
+            "description": "internal_error: An unexpected server error occurred."
+          }
+        },
+        "security": [
+          {
+            "bearer_header_Authorization": []
```
djjuhasz, line 94: Huh, it looks like authorization is disabled for this endpoint. I don't think that's intentional, is it?

internal/dips/api/design/design.go
```diff
@@ -0,0 +1,197 @@
...
+	Description("NotFoundProblem represents a missing resource problem.")
+	problemDetailsType("Not Found", "The requested DIP does not exist.", StatusNotFound)
+})
+
+var InternalProblem = Type("InternalProblem", func() {
+	Description("InternalProblem represents an unexpected server problem.")
+	problemDetailsType("Internal Server Error", "The DIP request could not be processed.", StatusInternalServerError)
+})
+
+var CreateDIPResult = Type("CreateDIPResult", func() {
```
djjuhasz: I think it's better to put the result definition directly in the response definition to avoid using a "#ref" in the generated schema.  For types that are only referenced once using a reference makes it harder to read the schema.

internal/dips/api/design/design.go
```diff
@@ -0,0 +1,197 @@
...
+	Description("CreateDIPResult contains the result of a DIP creation request.")
+	Field(1, "id", DIPID, "The id field contains the identifier assigned to the DIP.")
+	Required("id")
+})
+
+var DIP = Type("DIP", func() {
```
djjuhasz: I think the generated schema is more readable with this definition made directly in the result definition too.

internal/dips/api/gen/http/openapi3.json
```diff
@@ -0,0 +1,365 @@
...
+    },
+    "securitySchemes": {
+      "bearer_header_Authorization": {
+        "description": "Authenticate with an access token supplied as Authorization: Bearer [token]. Clients obtain tokens from an external OIDC provider.",
+        "scheme": "bearer",
+        "type": "http"
+      }
+    }
+  },
+  "info": {
```
djjuhasz, line 3: The alphabetical sorting of the API spec is new and makes it harder to read imho.  I would expect fields like "info" and "openapi" to be at the top of the spec.

internal/dips/api/design/design.go
```diff
@@ -0,0 +1,197 @@
...
+	// without adding an ErrorName discriminator to the RFC 9457 body.
+	//
+	// Goa initializes every named Type with an openapi:typename matching the Go
+	// type name. These error types should not become separate OpenAPI schemas:
+	// they only exist to give Goa distinct errors and examples. Replace Goa's
+	// generated typename so every response references the shared ProblemDetails
+	// schema instead. RemoveMeta is required because Meta appends values and the
+	// OpenAPI generator reads the first openapi:typename value.
+	RemoveMeta("openapi:typename")
+	Meta("openapi:typename", "ProblemDetails")
```
djjuhasz: Nice! It's unfortunate that you have to use this hack, but I definitely prefer the way the schema is rendered. 👌

djjuhasz (changes requested):
Thanks for the updates @jraddaoui.  I've added some more notes.

internal/dips/api/design/design.go
djjuhasz: You're right, I do strongly prefer that we don't use relative type URIs, so thanks for figuring out a way to support the "about:blank" type.  I agree that we don't need to extend the "Bad Request" response shape for this API.

Just to explain myself a bit, a big part of the reason I did c19f9559a1f610e277158919f8a4a37891c23f21 was so I could learn what problems you were encountering with the Goa generation of the `ProblemDetails` schema, and it was really helpful for me to better understand the Goa design DSL and its limitations.

.gitattributes
```diff
@@ -0,0 +1,3 @@
+hack/apis-mock/internal/gen/** linguist-generated
```
djjuhasz, line 1: Huh, I wonder what this does?

internal/dips/api/design/design.go
jraddaoui: Cool, and I think the learning will have its benefits in Enduro when we improve the error responses there. Also, rare, I didn't get an email or saw your previous update when I wrote mine last night :sweat_smile:

internal/dips/api/gen/http/openapi.json
jraddaoui: :man_shrugging:

internal/dips/api/design/design.go
jraddaoui: Well, it's just the description and I thought using those default names from Goa would help with the implementation, but they don't. There is also no way to remove that name from the final description, so I'll follow your suggestion.

jraddaoui: I have changed the names:

- not_valid -> bad_request
- internal_error -> internal_server_error

And removed the descriptions entirely to avoid duplication, these leaves the default description with the changed name:

- bad_request: Bad Request response.
- internal_server_error: Internal Server Error response.

Our description is only included in the example detail.

internal/dips/api/gen/http/openapi3.json
jraddaoui: See the comment above.

internal/dips/api/gen/http/openapi3.json
jraddaoui: This means it requires the security scheme named `bearer_header_Authorization`. The empty array means no scopes are required. I think this API will have authn, but not authz.

internal/dips/api/design/design.go
jraddaoui: It looks like even if we don't have the type and we put the attributes directly in the `Result()`, Goa still generates the component schema for each response and uses `$ref`.

I have removed the types and descriptions in the design and the refs are now named "CreateResponseBody" and "ShowResponseBody", instead of "CreateDIPResult" and "DIP". Let me know your preference.

internal/dips/api/design/design.go
jraddaoui: See previous comment.

internal/dips/api/gen/http/openapi3.json
jraddaoui: This was our fault, we were sorting when formatting with `jq`. I have removed that now in a different commit, so you can still read the feedback diff.

.gitattributes
jraddaoui: This is for Github to hide the diff on those files.

jraddaoui:
Thanks @djjuhasz, great catches! Let me know how it looks now ;)

jraddaoui:
Pretty basic WIP implementation: https://github.com/artefactual-sdps/preprocessing-sfa/pull/219

internal/dips/api/design/design.go
djjuhasz: Huh, it's surprising that Goa still uses `$ref` in the generated schema.  I'd leave it like you have it now.

.gitattributes
djjuhasz: Ah, that's cool. :smile:

djjuhasz (approved):
Looks great @jraddaoui! Thanks for doing all those updates. :sweat_smile:

jraddaoui:
After testing the initial implementation in #219, it looks like Goa response errors get in the way:

A request missing the authorization header:

```
curl -i \
  -H "Content-Type: application/json" \
  -d '{"docKey":"CH-000001"}' \
  http://localhost:8082/dips
    
HTTP/1.1 400 Bad Request
Content-Type: application/json
Date: Wed, 03 Jun 2026 19:16:02 GMT
Content-Length: 134

{
  "name": "missing_field",
  "id": "crlfBO-B",
  "message": "\"token\" is missing from header",
  "temporary": false,
  "timeout": false,
  "fault": false
}
```

A request with an unauthorized token:

```
curl -i \
  -H "Authorization: Bearer unauthorized" \
  -H "Content-Type: application/json" \
  -d '{"docKey":"CH-000001"}' \
  http://localhost:8082/dips

HTTP/1.1 401 Unauthorized
Content-Type: application/problem+json
Goa-Error: unauthorized
Date: Wed, 03 Jun 2026 19:19:48 GMT
Content-Length: 110

{
  "type": "about:blank",
  "title": "Unauthorized",
  "detail": "The bearer token is missing or invalid.",
  "status": 401
}
```

Similar, a request with an empty document key:

```
curl -i \
  -H "Authorization: Bearer test" \
  -H "Content-Type: application/json" \
  -d '{"docKey":""}' \
  http://localhost:8082/dips

HTTP/1.1 400 Bad Request
Content-Type: application/json
Date: Wed, 03 Jun 2026 19:23:03 GMT
Content-Length: 183

{
  "name": "invalid_length",
  "id": "4092IFvU",
  "message": "length of body.docKey must be greater or equal than 1 but got value \"\" (len=0)",
  "temporary": false,
  "timeout": false,
  "fault": false
}
```

But then, an error returned from the implementation:

```
curl -i \
  -H "Authorization: Bearer test" \
  -H "Content-Type: application/json" \
  -d '{"docKey":"error-doc-key"}' \
  http://localhost:8082/dips

HTTP/1.1 500 Internal Server Error
Content-Type: application/problem+json
Goa-Error: internal_server_error
Date: Wed, 03 Jun 2026 19:24:43 GMT
Content-Length: 116

{
  "type": "about:blank",
  "title": "Internal Server Error",
  "detail": "Failed to create DIP in the database",
  "status": 500
}
```

Also, an invalid UUID:

```
curl -i \
    -H "Authorization: Bearer test" \
    http://localhost:8082/dips/invalid-uuid

HTTP/1.1 400 Bad Request
Content-Type: application/json
Date: Wed, 03 Jun 2026 19:29:14 GMT
Content-Length: 209

{
  "name": "invalid_format",
  "id": "iKX6jbTk",
  "message": "id must be formatted as a uuid but got value \"invalid-uuid\", uuid: invalid-uuid: invalid UUID length: 12",
  "temporary": false,
  "timeout": false,
  "fault": false
}
```

Against a not found error returned in the implementation:

```
curl -i \
    -H "Authorization: Bearer test" \
    http://localhost:8082/dips/e8d32bd5-faa4-4ce1-bb50-55d9c28b306d

HTTP/1.1 404 Not Found
Content-Type: application/problem+json
Goa-Error: not_found
Date: Wed, 03 Jun 2026 19:31:04 GMT
Content-Length: 100

{
  "type": "about:blank",
  "title": "Not Found",
  "detail": "The requested DIP was not found.",
  "status": 404
}
```

I'm looking at two options:

- Add a middleware to transform Goa error responses to RFC 9457 problem details responses.
- Go back to default Goa error responses, ideally changing the content type value in the spec.

jraddaoui:
I tried the middleware approach as a proof of concept: f7f66f7.

The idea was to keep Goa validation in place, but normalize Goa validation errors into RFC 9457 Problem Details responses at the HTTP boundary. This made errors raised by Goa decoding/validation use the same response shape as errors returned by the service implementation.

This commit was fully AI generated by Codex (GPT-5.5 xhigh), so it should be reviewed carefully before we decide to follow this approach. I treated it only as a PoC to understand whether the middleware approach is viable.

Here are the response examples from that implementation:


A request missing the authorization header:

```bash
curl -i \
  -H "Content-Type: application/json" \
  -d '{"docKey":"CH-000001"}' \
  http://localhost:8082/dips

HTTP/1.1 400 Bad Request
Content-Type: application/problem+json
Goa-Error: missing_field
Date: Wed, 03 Jun 2026 20:33:14 GMT
Content-Length: 102

{
  "type": "about:blank",
  "title": "Bad Request",
  "detail": "\"token\" is missing from header",
  "status": 400
}
```

A request with an unauthorized token:

```bash
curl -i \
  -H "Authorization: Bearer unauthorized" \
  -H "Content-Type: application/json" \
  -d '{"docKey":"CH-000001"}' \
  http://localhost:8082/dips

HTTP/1.1 401 Unauthorized
Content-Type: application/problem+json
Goa-Error: unauthorized
Date: Wed, 03 Jun 2026 20:33:14 GMT
Content-Length: 110

{
  "type": "about:blank",
  "title": "Unauthorized",
  "detail": "The bearer token is missing or invalid.",
  "status": 401
}
```

A request with an empty document key:

```bash
curl -i \
  -H "Authorization: Bearer test" \
  -H "Content-Type: application/json" \
  -d '{"docKey":""}' \
  http://localhost:8082/dips

HTTP/1.1 400 Bad Request
Content-Type: application/problem+json
Goa-Error: invalid_length
Date: Wed, 03 Jun 2026 20:33:14 GMT
Content-Length: 150

{
  "type": "about:blank",
  "title": "Bad Request",
  "detail": "length of body.docKey must be greater or equal than 1 but got value \"\" (len=0)",
  "status": 400
}
```

An error returned from the implementation:

```bash
curl -i \
  -H "Authorization: Bearer test" \
  -H "Content-Type: application/json" \
  -d '{"docKey":"error-doc-key"}' \
  http://localhost:8082/dips

HTTP/1.1 500 Internal Server Error
Content-Type: application/problem+json
Goa-Error: internal_server_error
Date: Wed, 03 Jun 2026 20:33:24 GMT
Content-Length: 116

{
  "type": "about:blank",
  "title": "Internal Server Error",
  "detail": "Failed to create DIP in the database",
  "status": 500
}
```

An invalid UUID:

```bash
curl -i \
  -H "Authorization: Bearer test" \
  http://localhost:8082/dips/invalid-uuid

HTTP/1.1 400 Bad Request
Content-Type: application/problem+json
Goa-Error: invalid_format
Date: Wed, 03 Jun 2026 20:33:24 GMT
Content-Length: 176

{
  "type": "about:blank",
  "title": "Bad Request",
  "detail": "id must be formatted as a uuid but got value \"invalid-uuid\", uuid: invalid-uuid: invalid UUID length: 12",
  "status": 400
}
```

A not found error returned from the implementation:

```bash
curl -i \
  -H "Authorization: Bearer test" \
  http://localhost:8082/dips/e8d32bd5-faa4-4ce1-bb50-55d9c28b306d

HTTP/1.1 404 Not Found
Content-Type: application/problem+json
Goa-Error: not_found
Date: Wed, 03 Jun 2026 20:33:24 GMT
Content-Length: 100

{
  "type": "about:blank",
  "title": "Not Found",
  "detail": "The requested DIP was not found.",
  "status": 404
}
```

jraddaoui:
I also tested the option of going back to Goa's default error responses.

With this approach, the API uses Goa's standard error schema consistently for both Goa validation errors and errors returned by the service implementation. I was also able to update the design so the OpenAPI schema shows `application/json` for these error responses, matching the content type actually returned by the server.

I kept the commit history in #219, so both implementations are available there, and the relevant design commits are also included in this PR.

Here are the response examples from that design and implementation:


A request missing the authorization header:

```bash
curl -i \
  -H "Content-Type: application/json" \
  -d '{"docKey":"CH-000001"}' \
  http://localhost:8082/dips

HTTP/1.1 400 Bad Request
Content-Type: application/json
Date: Wed, 03 Jun 2026 20:22:47 GMT
Content-Length: 134

{
  "name": "missing_field",
  "id": "atghjlsg",
  "message": "\"token\" is missing from header",
  "temporary": false,
  "timeout": false,
  "fault": false
}
```

A request with an unauthorized token:

```bash
curl -i \
  -H "Authorization: Bearer unauthorized" \
  -H "Content-Type: application/json" \
  -d '{"docKey":"CH-000001"}' \
  http://localhost:8082/dips

HTTP/1.1 401 Unauthorized
Content-Type: application/json
Goa-Error: unauthorized
Date: Wed, 03 Jun 2026 20:22:47 GMT
Content-Length: 140

{
  "name": "unauthorized",
  "id": "YsROtcdy",
  "message": "the bearer token is missing or invalid",
  "temporary": false,
  "timeout": false,
  "fault": false
}
```

A request with an empty document key:

```bash
curl -i \
  -H "Authorization: Bearer test" \
  -H "Content-Type: application/json" \
  -d '{"docKey":""}' \
  http://localhost:8082/dips

HTTP/1.1 400 Bad Request
Content-Type: application/json
Date: Wed, 03 Jun 2026 20:23:09 GMT
Content-Length: 183

{
  "name": "invalid_length",
  "id": "qlo7nUL3",
  "message": "length of body.docKey must be greater or equal than 1 but got value \"\" (len=0)",
  "temporary": false,
  "timeout": false,
  "fault": false
}
```

An error returned from the implementation:

```bash
curl -i \
  -H "Authorization: Bearer test" \
  -H "Content-Type: application/json" \
  -d '{"docKey":"error-doc-key"}' \
  http://localhost:8082/dips

HTTP/1.1 500 Internal Server Error
Content-Type: application/json
Goa-Error: internal_server_error
Date: Wed, 03 Jun 2026 20:23:09 GMT
Content-Length: 145

{
  "name": "internal_server_error",
  "id": "XZipMB1V",
  "message": "failed to create DIP in the database",
  "temporary": false,
  "timeout": false,
  "fault": true
}
```

An invalid UUID:

```bash
curl -i \
  -H "Authorization: Bearer test" \
  http://localhost:8082/dips/invalid-uuid

HTTP/1.1 400 Bad Request
Content-Type: application/json
Date: Wed, 03 Jun 2026 20:23:09 GMT
Content-Length: 209

{
  "name": "invalid_format",
  "id": "Yt0Ycmmd",
  "message": "id must be formatted as a uuid but got value \"invalid-uuid\", uuid: invalid-uuid: invalid UUID length: 12",
  "temporary": false,
  "timeout": false,
  "fault": false
}
```

A not found error returned from the implementation:

```bash
curl -i \
  -H "Authorization: Bearer test" \
  http://localhost:8082/dips/e8d32bd5-faa4-4ce1-bb50-55d9c28b306d

HTTP/1.1 404 Not Found
Content-Type: application/json
Goa-Error: not_found
Date: Wed, 03 Jun 2026 20:23:09 GMT
Content-Length: 130

{
  "name": "not_found",
  "id": "8Ve0_88Q",
  "message": "the requested DIP was not found",
  "temporary": false,
  "timeout": false,
  "fault": false
}
```


**One caveat:** the `Goa-Error` response header is not present on every error response. Goa adds it for named service errors returned by the implementation, but validation/decode errors produced by the Goa transport layer do not include it. The error name is always present in the JSON response body, so clients should use the body's `name` field rather than relying on the header. I'm leaving the header behavior as Goa generates it instead of adding custom middleware to normalize or remove it.

jraddaoui:
My take is that we should keep the current PR state, using the default Goa error responses unless the client specifically asks for RFC 9457 Problem Details.

We have a Problem Details proof of concept, but aligning Goa validation errors, implementation errors, examples, and schema generation adds extra design and runtime complexity that does not seem justified without a concrete requirement.

With the error response content type updated to match the actual server responses, and with the generated schema representing the default Goa error response body, the default Goa error schema looks like a reasonable first approach. I’ll squash the PR before merging, but keep an unsquashed branch version for reference.

djjuhasz:
@jraddaoui I'm fine with using the default Goa error responses as you've recommended.

I don't love that our API error response shape is predefined by Goa, but I agree that it's not worth the extra complexity of adding middleware or fighting with the default Goa schema generation, error validation and  error handling. I think Goa's error response schema is sufficient, though it isn't what I would choose if we had the freedom to easily define our own schema.  I'm glad you found a way to have a standard `Content-Type application-json` in the generated OpenAPI schemas instead of a custom Goa content type.

I think the worst case scenario is that we choose to ditch Goa at some point in the future and have to create a v2 (or v3 or whatever) API specification that doesn't use the Goa error responses, and that's an acceptable exit plan in my opinion.

jraddaoui:
Branch with the full commit history, in case we want to go back to RFC 9457: https://github.com/artefactual-sdps/preprocessing-sfa/tree/dev/issue-216-dips-api-design-history.

Issue #222, opened 2026-06-12
Add separate workflow for third parties


sallain:
**Is your feature request related to a problem? Please describe.**

SFA uses Enduro for their own workflow, but they also offer Enduro as a service for third-party clients. The third parties do not use the APIS->AIS integration, and instead just need the old post-storage metadata workflow.

**Describe the solution you'd like**

In discussions, we have decided to create a separate workflow that can be deployed for the third parties. This workflow should not include any of the APIS work.

**Describe alternatives you've considered**

**Additional context**

sallain:
I believe that this should NOT include the DIP work as well, but I will check in with Iva.

jraddaoui:
My understanding is that they will use a separate Enduro instance for the third-party clients, including a separate instance of their custom child workflows worker.

I can fully restore the previous poststorage workflow as a third workflow registered in their custom worker. They already have an APIS enabled flag in their configuration, so we could use that flag to decide which poststorage workflow is registered. Then each Enduro instance can be configured to run the matching workflow.

That would leave three workflows available in their custom `sfa-enduro-worker`:

- `preprocessing`: APIS integration enabled or disabled based on the configuration flag.
- `poststorage-apis`: current APIS integration workflow, registered when the APIS integration is enabled.
- `poststorage-cantons`: previous AIS bundle workflow, registered when the APIS integration is disabled.

For the SFA instance:

- Custom worker: APIS integration enabled in `preprocessing`, so `poststorage-apis` is registered.
- Enduro: runs `preprocessing` and `poststorage-apis`.

For the third-party instance:

- Custom worker: APIS integration disabled in `preprocessing`, so `poststorage-cantons` is registered.
- Enduro: runs `preprocessing` and `poststorage-cantons`.

The DIP workflow lives in a different worker, so it should not affect any of the existing workflows. But I assume that, without the APIS integration, they will not support DIP creation for third-party clients.

sallain:
That all sounds good to me @jraddaoui. I'll send this comment along to SFA to ensure we've got it right.

sallain:
@jraddaoui The client has confirmed that this looks good.

jraddaoui:
The PR above restored the previous AIS bundle poststorage workflow, now called `poststorage-cantons`.

However, when we initially developed it, Enduro was not collecting tasks from poststorage child workflows. So, while the APIS poststorage workflow will show tasks in the Enduro API/UI, this one still does not.

PR #223, opened 2026-06-16
Restore legacy poststorage workflow

jraddaoui:
Restore the previous poststorage workflow as poststorage-cantons while keeping the APIS based workflow available as poststorage-apis. Register one or the other based on the APIS configuration flag.

Refs #222.


cmd/worker/workercmd/cmd.go
```diff
@@ -285,6 +303,36 @@ func (m *Main) registerPoststorageWorkflow(packages *ssclient.PackagesService, a
...
+		m.temporalWorker.RegisterActivityWithOptions(
+			archivezip.New().Execute,
+			temporalsdk_activity.RegisterOptions{Name: archivezip.Name},
+		)
+		m.temporalWorker.RegisterActivityWithOptions(
+			bucketupload.New(cantonsBucket).Execute,
+			temporalsdk_activity.RegisterOptions{Name: bucketupload.Name},
+		)
+
+		return
```
djjuhasz: This return is easy to miss, I missed it on my first read.  A possible alternative is to create `m.registerAPIs()` and `m.registerCantons()` functions and then do a simple `if ... else` conditional to determine which registrations are called.  It's up to you though, this isn't a blocker.

internal/workflows/poststorage_apis.go
```diff
@@ -19,19 +19,15 @@ import (
 	"github.com/artefactual-sdps/sfa-enduro-workflows/internal/config"
 )
 
-type Poststorage struct {
-	cfg         config.PoststorageConfig
-	apisEnabled bool
+type PoststorageAPIS struct {
```
djjuhasz, line 22: I think it's better to rename the file `poststorage_apis.go` to match the type.

internal/workflows/poststorage_test.go
```diff
@@ -94,7 +94,7 @@ func (s *TestSuite) setup(cfg *config.PoststorageConfig, apisEnabled bool) {
 		temporalsdk_activity.RegisterOptions{Name: apis.PollImportRunStatusActivityName},
 	)
 
-	s.workflow = workflows.NewPoststorage(*cfg, apisEnabled)
+	s.workflow = workflows.NewPoststorageAPIS(*cfg)
 }
 
 func TestPoststorage(t *testing.T) {
```
djjuhasz: I'd rename this `TestPoststorageAPIS` to match the type name, and rename this file `poststorage_apis_test.go` too.

djjuhasz (approved):
LGTM @jraddaoui. I had a few suggestions, but none of them are blockers.

cmd/worker/workercmd/cmd.go
jraddaoui: I have used if/else inside the existing function and added some comments to avoid duplication registering activities. I hope that's better for you too ;)

PR #225, opened 2026-06-18
Pass child workflow user email to APIS

jraddaoui:
Use the user email supplied by Enduro child workflow params when
creating APIS import tasks and runs. Keep the existing "sfa-enduro"
fallback when no user is provided.

Refs #173, #174 and https://github.com/artefactual-sdps/enduro/issues/1666.


djjuhasz (approved):
LGTM. 👍

PR #228, opened 2026-06-23
Handle APIS default values failure

jraddaoui:
Detect the APIS 400 response used when default metadata values cannot be determined, map it to a content failure in preprocessing, and update the "Submit metadata to APIS" task error message.

Refs #173.


djjuhasz (approved):
LGTM 👍

PR #230, opened 2026-06-25
Use a shared bagit-gython validator

djjuhasz:
Revert commit 67f3aac because bagit-gython now (v0.5.0) supports validating multiple bags concurrently via the validator API.


cmd/worker/workercmd/cmd.go
```diff
@@ -78,6 +79,18 @@ func (m *Main) Run(ctx context.Context) error {
 
 	veraPDFValidator := fvalidate.NewVeraPDFValidator(m.cfg.Preprocessing.FileValidate.VeraPDF.Path)
 
+	// Set up BagIt validator.
+	bagValidator, err := bagit.NewValidator(bagit.WithPoolSize(1))
```
jraddaoui: See https://github.com/artefactual-sdps/enduro/pull/1683#discussion_r3471184259.

jraddaoui (approved):
Thanks!

djjuhasz:
I added configuration for bagit_gython, similar to https://github.com/artefactual-sdps/enduro/commit/09a36d67ae23fd6b04cdc8ba7ea5d8860a19b7d5

cmd/worker/workercmd/cmd.go
```diff
@@ -78,6 +79,21 @@ func (m *Main) Run(ctx context.Context) error {
...
+	)
+	if err != nil {
+		m.logger.Error(err, "Unable to create BagIt validator.")
+		return err
+	}
+	defer func() {
+		if err := bagValidator.Close(); err != nil {
+			m.logger.Error(err, "Error closing BagIt validator.")
+		}
+	}()
```
jraddaoui: I believe this function doesn't block and that's why we have the `Close()` method. We should add the validator to the main struct and close it there.

internal/config/config_test.go
```diff
@@ -536,6 +555,51 @@ token = "mock-token"
...
+username = "test"
+key = "test"
+[apis]
+enabled = true
+url = "http://apis.example.test"
+`,
+			wantFound: true,
+			wantErr: `invalid configuration
+Preprocessing.BagValidate: PoolSize: 0 is less than the minimum value (1)`,
+		},
```
jraddaoui: Not a blocker, but could we put this into an existing error case? To avoid writing the full config each time we test a single failure.

README.md
```diff
@@ -479,7 +483,7 @@ file) also uses a number of other more general Enduro
 
 * `archiveextract`
 * `bagcreate`
-* `bagvalidate`
+* `bagvalidator`
```
jraddaoui: Why this?

jraddaoui (changes requested):
Thanks @djjuhasz!

README.md
djjuhasz: Partially because I was trying to avoid the name collision with the imported `bagvalidate` package, but also because I realized package names are usually a noun, or a plural noun (e.g. strings, os, context).  There are exceptions (e.g. testing), but I think it's probably better to use a noun in most cases.

jraddaoui: But this is listing other activities from temporal-activities, and `bagvalidate` is the package name there:

https://github.com/artefactual-sdps/temporal-activities/tree/main/bagvalidate

djjuhasz: Ohhh, that was probably just a search & replace mistake. :flushed:

djjuhasz:
@jraddaoui this is ready for another look.  I'll squash the fix commits before merging.

jraddaoui (approved):
Thanks @djjuhasz!

PR #231, opened 2026-07-06
Minor and cosmetic fixes

jraddaoui:
- Fix test error assertions
- Avoid opening SIP manifest to write PREMIS objects
- Check remove error in SIP transformation
- Delete fformat fake MockValidator
- Avoid SFA specifics in fformat and fsutil packages
- Fix var naming in preprocessing workflow
- Make binary stripping configurable


djjuhasz (approved):
Looks great, thanks @jraddaoui. 👍

PR #237, opened 2026-07-20
Update DIPs API

djjuhasz:
Ref #216 

- Add `GET /livez` endpoint to check service liveness
- Add optional `ignoreCache` field to `POST /dips` endpoint
- Add `error_message` field to `GET /dips/{uuid}` endpoint

### DIFF
--- a/internal/dips/api/design/design.go
+++ b/internal/dips/api/design/design.go
@@ -12,7 +12,7 @@ var BearerAuth = BearerSecurity("bearer", func() {
 var _ = API("DIPs", func() {
 	Title("DIPs API")
 	Description("The DIPs API is used to request DIP creation and retrieve DIP details.")
-	Version("0.1.0")
+	Version("0.2.0")
 })
 
 var DIPID = Type("DIPID", String, func() {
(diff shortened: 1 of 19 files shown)


internal/dips/api/design/design.go
sevein, line 53: The errors remain service-scoped, not sure if this is intentional. You may have to move these declarations into the methods that actually use them.

internal/dips/api/design/design.go
```diff
@@ -95,16 +110,20 @@ var _ = Service("DIPs", func() {
 			Field(1, "id", DIPID, "The id field uniquely identifies the DIP.")
 			Field(2, "docKey", DocKey, "The docKey field contains the document key used to create the DIP.")
 			Field(3, "status", DIPStatus, "The status field contains the current DIP status.")
-			Field(4, "created_at", DateTime, "The created_at field contains the time when the DIP was requested.")
-			Field(5, "started_at", DateTime, "The started_at field contains the time when DIP processing started.")
+			Field(4, "error_message", String, func() {
+				Description("The error_message field contains an error message if the DIP failed.")
+				Example("ACTAPro returned: 404 Not Found; Document not found")
+			})
```
sevein: We should eventually update this design to adopt `Attribute` instead of `Field`. `Field` is an alias for `Attribute` plus an `rpc:tag` useful when using Protobuf/gRPC, but there is no gRPC transport configured. Also, in gRPC world, the way `error_message` is added, renumbering existing tags, is a backward-incompatible change.

internal/dips/api/design/design.go
```diff
@@ -57,16 +57,28 @@ var _ = Service("DIPs", func() {
...
+
 	Method("create", func() {
 		Description("The create method requests DIP creation for a document key.")
 		Payload(func() {
 			BearerTokenField(1, "token", String, "The token field contains the OIDC bearer token.")
 			Field(2, "docKey", DocKey, "The docKey field contains the document key used to create the DIP.")
+			Field(3, "ignoreCache", Boolean, func() {
+				Description("The ignoreCache field indicates whether to ignore the cache.")
+				Example(true)
+			})
```
sevein: A couple of ideas:
* Should we use `Default(false)` so clients and the implementation have consistent behavior when `ignoreCache` is omitted, i.e., omitting it is equivalent to setting it to false?
* Could the description also clarify what constitutes a match? E.g. "When true, creates a new DIP even if a completed DIP for the same docKey is cached. When false or omitted, a matching cached DIP may be reused." or similar?

internal/dips/api/design/design.go
```diff
@@ -57,16 +57,28 @@ var _ = Service("DIPs", func() {
...
 
 	HTTP(func() {
-		Response("bad_request", StatusBadRequest, goaErrorResponse())
-		Response("unauthorized", StatusUnauthorized, goaErrorResponse())
 		Response("internal_server_error", StatusInternalServerError, goaErrorResponse())
 	})
 
+	Method("healthz", func() {
+		Meta("openapi:summary", "DIPs service health check")
+		Description("The healthz method checks the health of the DIPs service.")
```
sevein: The issue describes this as a liveness check rather than a readiness check. Could we make that explicit here? Should `/healthz` only confirm that the service is running, without checking dependencies such as AIS, AMSS, or the database? An alternative description could be:

> Reports whether the DIPs service is running. It does not check readiness or downstream dependencies.

Another option is to consider `/livez` and/or `/readyz`. Context:
* https://kubernetes.io/docs/reference/using-api/health-checks/
* https://goa.design/docs/3-ecosystem/clue/#the-health-package

internal/dips/api/design/design.go
djjuhasz: It looks like Goa intends for common errors to be defined at the API or service level, and then called by specific methods, see: https://goa.design/docs/1-goa/dsl-reference/#service-level-errors

The rendered Open API files have the errors correctly listed in the applicable method responses.

internal/dips/api/design/design.go
djjuhasz: I'll do that now.

internal/dips/api/design/design.go
djjuhasz: Of course there are `/livez` and `/readyz` status check conventions, I should have guessed. :laughing:

I've gone with `/livez` and fleshed out the method description.

djjuhasz:
@sevein ready for another look.

sevein (approved):
LGTM

PR #240, opened 2026-07-23
Allow files in the SIP content directory

djjuhasz:


internal/activities/validate_structure_test.go
```diff
@@ -229,12 +216,32 @@ func TestValidateStructure(t *testing.T) {
...
+		},
+		{
+			name: "Validates a digitized AIP without dossiers",
+			params: activities.ValidateStructureParams{
+				SIP: digitizedAIP(t,
+					fs.WithFile("00000001.jp2", ""),
+					fs.WithFile("00000001_PREMIS.xml", ""),
+					fs.WithFile("Prozess_Digitalisierung_PREMIS.xml", ""),
+				),
+			},
```
sevein, line 238: If digitized AIPs without a dossier are now considered valid, do we also need to update TransformSIP? ([entries[0].Name()](https://github.com/artefactual-sdps/sfa-enduro-workflows/blob/933d13805b58b9c47d455c95dae2a5cde548ff13/internal/activities/transform_sip.go#L49-L53) in transform_sip.go).

sevein (approved):
LGTM. I only have one question!

internal/activities/validate_structure_test.go
djjuhasz: The fix for issue #241 removes the transformation code that moves the "Prozess_Digitalisierung_PREMIS.xml" to the PIP metadata/ directory, so that particular reference to the dossier directory will be gone.

PR #242, opened 2026-08-04
Don't move Prozess_Digitalisierung_PREMIS.xml

### LINKED ISSUE #241: Problem: identifier for Prozess_Digitalisierung_PREMIS.xml doesn't end up in METS
**Describe the bug**

The *Create identifier.json* copies identifiers from the metadata manifests (either metadata.xml or UpdatedAreldaMetadata.xml) into an identifier.json file. Archivematica later consumes this file to insert those identifiers into the AIP METS. This works as expected for files that remain in the objects directory.

However, _Prozess_Digitalisierung_PREMIS.xml_ gets moved to the metadata directory, and somewhere along the way (while in Archivematica, I think) the system loses track of the connection between the file and its identifier. My hypothesis is that the path for this file in the identifiers.json, `metadata/Prozess_Digitalisierung_PREMIS.xml`, is no longer correct by the time that Archivematica's **Job: Load persistent identifiers from external file** actually runs.

This only applies to DigitizedSIPs

**To Reproduce**

1. Ingest a DigitizedSIP
2. Download the AIP
3. Open the identifier.json file -- `data/objects/metadata` then drill down until you find it
4. Check the entry for `Prozess_Digitalisierung_PREMIS.xml` and copy its identifier
5. Open the AIP METS file and search for the identifier

**Expected behavior**

The identifier for `Prozess_Digitalisierung_PREMIS.xml` should appear in a premis:objectIdentifier field like this:

```
            <premis:objectIdentifier>
              <premis:objectIdentifierType>local</premis:objectIdentifierType>
              <premis:objectIdentifierValue>_zah8zif5OongoN7aek6eoda</premis:objectIdentifierValue>
            </premis:objectIdentifier>
```

However, it's only listed in the ExifTool output (XMP fields). The identifiers for all other files in the identifier.json field should appear in the correct premis:ObjectIdentifier field.

**Screenshots**

**Additional context**

djjuhasz:
Fixes #241.

- Don't move the Prozess_Digitalisierung_PREMIS.xml file to the PIP metadata directory when transforming the SIP so it's treated as a preservation object and its identifier is preserved in the METS file
- Change the Prozess_Digitalisierung_PREMIS.xml path in the "identifiers.json" file to match its path in the PIP content directory

### DIFF
--- a/internal/activities/transform_sip.go
+++ b/internal/activities/transform_sip.go
@@ -8,7 +8,6 @@ import (
 
 	"go.artefactual.dev/tools/fsutil"
 
-	"github.com/artefactual-sdps/sfa-enduro-workflows/internal/enums"
 	"github.com/artefactual-sdps/sfa-enduro-workflows/internal/pips"
 	"github.com/artefactual-sdps/sfa-enduro-workflows/internal/sip"
 )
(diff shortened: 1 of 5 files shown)


jraddaoui (approved):
LGTM!

